PR #1, opened 2026-09-17
feat: run thunderstorm from cloud sessions

desertthunder:
## What

Makes the thunderstorm workflow runnable from a cloud session, and repairs the CI it runs against. Board operations have a documented path through the GitHub MCP tools alongside `gh`; label definitions reach GitHub through a workflow rather than a local `gh`; a SessionStart hook warms the caches a web container starts without and tracks the stable toolchain; commit messages are checked by a git hook and reported in CI; and `.github/pull_request_template.md` puts the body structure where GitHub can find it.

## Why

Nothing in the repository mentioned cloud sessions, and six skills plus `sync-labels.sh` called `gh` unconditionally, so a web session could not claim an issue, move a status, or apply `.github/labels.yml` at all. Two transport differences are written down because they silently corrupt the board: the MCP issue update replaces an issue's whole `labels` and `assignees` array where `gh issue edit` edits only what it names, and label definitions do not go through MCP.

CI was already red on `edge` ([run 56](https://github.com/stormlightlabs/thunderus/actions/runs/35235635216)) before this branch existed, and green CI is a precondition for feature work, so those failures are fixed here:

| Failure | Cause |
| --- | --- |
| Dependency policy | RUSTSEC advisory on `rustls 0.23.41` via `ureq`; now 0.23.45 |
| Clippy | 13 redundant `use helpers::*` imports, rejected by newer stable |
| `cargo fmt` | doubled blank lines left by removing those imports |
| `url_reader` | test agent lacked production's `.proxy(None)`, so the redirect guard it asserts never ran wherever a proxy existed |
| `headless` cancel | `assert!(stdout.is_empty())` contradicted its own fixture, which streams a line before waiting for the cancel |

The last two were latent defects rather than noise: one test passed without exercising its subject, the other passed only by winning a race that neither a runner nor a container reliably wins.

## Verification

Run against rustc 1.98.1, the same stable CI installs:

- `cargo fmt --all -- --check`, and `cargo clippy --workspace --all-targets --all-features --locked -- -D warnings` — both clean.
- `cargo test --workspace --all-features --locked` — 1846 passed, 0 failed, across 7 targets.
- `cargo doc --workspace --all-features --no-deps --locked` under `RUSTDOCFLAGS=-D warnings`, and `cargo package -p thndrs-agent --locked` — clean.
- `pnpm --dir docs build` — 92 pages, all internal links valid.
- `.claude/hooks/session-start.sh` — exit 0, no stdout, idempotent; exits silently with `CLAUDE_CODE_REMOTE` unset.
- `sync-labels.sh` against a stubbed `gh`: 30/30 labels parse; a label carried only by a pull request is skipped rather than deleted; an unreadable count refuses every deletion.

## Not covered

- The labels workflow has only ever run its dry run. `workflow_dispatch` needs the file on the default branch, so merging this is what enables the apply path.
- Four commit messages predate the checker and carry warnings, left rather than rewriting pushed history.
- The label set is unapplied and the repository has no issues, so no thunderstorm run has been exercised end to end.
- CI still floats on `stable` while the MSRV job pins 1.88. The container now tracks stable to match; pinning both to an exact version is the alternative, and would need `cargo +1.88` overrides in the MSRV job.
- No changelog entry: this changes contributor workflow, not `thndrs` behavior.


crates/thndrs/src/core/url_reader.rs
```diff
@@ -585,6 +585,10 @@ mod tests {
     fn test_agent(timeout: Duration, resolver: impl Resolver) -> ureq::Agent {
         let config = ureq::Agent::config_builder()
             .max_redirects(0)
+            // Mirrors public_fetch_agent. Without it, a proxy in the
+            // environment resolves the destination itself, the test's resolver
```
desertthunder: this could be a rustdoc comment

internal/thunderstorm.md
```diff
@@ -20,6 +20,38 @@ A run ends when every sub-issue reaches a terminal state, the budget is spent,
 or a worker escalates. A worker that finds adjacent work files a new issue and
 does not start it in this run.
 
+## Where a run happens
+
+A run works from a local checkout or from a cloud session on the web. The
+protocol is the same; the transport to GitHub is not.
+
+| Host          | GitHub through   | Worktrees                          |
```
desertthunder: markdown should be formatted with prettier

desertthunder (commented):
A few small changes

crates/thndrs/src/core/url_reader.rs
desertthunder: Agreed, and it belonged on the function rather than in the middle of the builder chain. Moved to a `///` doc on `test_agent` in 4793ca0, with the reason stated in full: `proxy(None)` is what makes the redirect guard testable at all, because a proxy resolves the destination itself and the resolver passed in never runs.

Verified after the change: `cargo fmt --all -- --check` clean, the five `url_reader` tests pass, and `cargo doc -p thndrs --all-features --no-deps --locked` builds clean under `RUSTDOCFLAGS=-D warnings`.

One caveat on that last point — the doc sits inside a `#[cfg(test)]` module, so rustdoc does not process it and the intra-doc link to `public_fetch_agent` is not actually validated by that build. It resolves by eye, but strict rustdoc is not the thing confirming it.

— claude-opus-5 · high

---

internal/thunderstorm.md
desertthunder: Done in 6679d6d. Ran `prettier` over the ten markdown files this branch touches; five differed and are now clean under `prettier --check`.

The change is table alignment only — padding and separator widths. Prose is untouched because `proseWrap` defaults to `preserve`, so existing line breaks survive. That also pulled a few tables predating this branch back into alignment, notably the Commands and Skills tables in `internal/thunderstorm.md`.

One thing worth deciding separately: `docs/.prettierrc` governs `docs/` only, so everything under `internal/` and `.claude/` formats with prettier's defaults rather than your 120-column settings. Right now that difference is invisible because prettier does not rewrap prose, but it would bite the moment anyone sets `proseWrap`. A root `.prettierrc` would make the convention explicit, and a `--check` step in CI would keep it from drifting again the way these tables did. I have not done either here — it is beyond what this PR was for, and the config choice is yours.

— claude-opus-5 · high

---

.github/workflows/labels.yml
```diff
@@ -0,0 +1,77 @@
...
+    name: Sync the label set
+    if: github.event_name == 'workflow_dispatch'
+    runs-on: ubuntu-latest
+    permissions:
+      contents: read
+      # The narrowest scope GitHub offers for labels. It also permits editing
+      # and commenting on issues; there is no labels-only scope.
+      issues: write
+    steps:
+      - uses: actions/checkout@v4
```
desertthunder, line 73: medium · .github/workflows/labels.yml:62 — the `apply` job checks out whatever `ref` the dispatch names and runs that branch's copy of `sync-labels.sh` with `issues: write` → the isolation the comment at lines 27-31 describes only covers the `pull_request` trigger; `github-board` tells a cloud session to dispatch this workflow itself, and `internal/thunderstorm.md:211` says the `ref` "may then name any branch holding the file", so an agent can push a changed script to `agent/<n>`, dispatch on that ref, and run it against every issue and label before a human has read the diff → pin `actions/checkout` here to `ref: edge`, gate the job with `if: github.ref == 'refs/heads/edge'`, and have the skill and `thunderstorm.md` say dispatch targets `edge` only. The workflow file itself also comes from the dispatch ref, so a hard guarantee needs an `environment` with required reviewers on this job.

— claude-fable-5-1 · default

---

.githooks/commit-msg
```diff
@@ -0,0 +1,12 @@
...
+#
+# Enable it once per checkout:
+#
+#   git config core.hooksPath .githooks
+#
+# CI runs the same checker over a pull request's commits, so a checkout without
+# the hook is caught later rather than never.
+set -euo pipefail
+
+exec "$(git rev-parse --show-toplevel)/.claude/scripts/check-commit-message.py" "$1"
```
desertthunder: medium · .githooks/commit-msg:12 — the hook grades the whole `COMMIT_EDITMSG`, including the diff `git commit -v` (or `commit.verbose=true`) appends below the scissors line → git strips that diff after the `commit-msg` hook runs, so a verbose commit whose diff carries a >72-column line containing a space is rejected for content that never enters the message; reproduced: a one-line `feat:` message over a diff adding a 125-character line fails with `line 9 is 125 characters, over the 72 allowed` → in `check()`, stop reading at the scissors line (`# ------------------------ >8 ------------------------`, honoring `core.commentChar`) before grading.

medium · .githooks/commit-msg:12 — git runs `commit-msg` for `git merge` as well, and `Merge branch 'agent/x' into edge` fails the subject regex → a local `git merge --no-ff`, or merging `edge` into an agent branch to resolve a conflict, stops with `Not committing merge; use 'git commit' to complete the merge` (reproduced), and the only way through is `--no-verify`, the same flag that skips the check for the next real message → skip when `MERGE_HEAD` exists in the git dir, or accept the `Merge ` and `Revert "` subjects git generates.

— claude-fable-5-1 · default

---

.claude/scripts/check-commit-message.py
```diff
@@ -0,0 +1,146 @@
...
+        # %x00 separates commits; a message may contain any other byte.
+        out = subprocess.run(
+            ["git", "log", "--format=%H%n%B%x00", argv[1]],
+            capture_output=True,
+            text=True,
+            check=True,
+        ).stdout
+        entries = [entry for entry in out.split("\0") if entry.strip()]
+    else:
+        with open(argv[0], encoding="utf-8") as handle:
```
desertthunder: medium · .claude/scripts/check-commit-message.py:102 — non-UTF-8 input crashes the checker instead of producing a verdict → `open(..., encoding="utf-8")` raised `UnicodeDecodeError` on a `COMMIT_EDITMSG` holding `fix: caf\xe9` (reproduced: the committer gets a traceback and exit 1 with no rule named); the `--range` path at lines 94-99 decodes `git log` with the same strict `text=True` and adds `check=True`, so a base SHA the checkout cannot see raises `CalledProcessError` (reproduced with `deadbeef..HEAD`). Under `--warn` in the `commits` job, whose shell is `bash -eo pipefail`, either traceback fails the job that the comment above it says never fails → read bytes and decode with `errors="replace"`, and in `--warn` mode turn a `git log` failure into a printed warning with exit 0. I could not reproduce raw-byte storage for the range case because this sandbox's git transcodes messages to UTF-8 on commit; stock git stores the bytes and only warns.

— claude-fable-5-1 · default

---

.claude/scripts/check-commit-message.py
```diff
@@ -0,0 +1,146 @@
...
+        problems.append("no blank line between the subject and the body")
+
+    # A fenced block holds output or commands that wrapping would corrupt, and a
+    # line without spaces is a URL or a path that cannot be wrapped at all.
+    fenced = False
+    for number, line in enumerate(lines[1:], start=2):
+        if line.lstrip().startswith("```"):
+            fenced = not fenced
+            continue
+        if fenced or TRAILER.match(line) or " " not in line.strip():
```
desertthunder: low · .claude/scripts/check-commit-message.py:73 — `TRAILER` is `^[A-Za-z][A-Za-z-]*: .+$` and is applied to every body line, so any line beginning `Note: `, `See: `, or `Fixes: ` escapes the 72-column rule, and an unclosed ``` fence exempts everything after it → both reproduced with a 200-character line that passes clean, so the column rule the skill advertises has two silent exits → treat as trailers only the lines of the final paragraph, and report an unclosed fence as a violation.

low · .claude/scripts/check-commit-message.py:36 — the `#` comment strip also runs in `--range` mode, where `%B` is the stored message and carries no comments → a stored subject such as `#12 broke the build` is dropped and the next line is graded as the subject; reproduced: `#12 broke\nfeat: whatever` passes clean → strip comments only in file mode, and only with the character `core.commentChar` names.

— claude-fable-5-1 · default

---

.claude/skills/github-board/SKILL.md
```diff
@@ -50,6 +90,15 @@ gh issue edit <n> --add-assignee @me
 gh issue edit <n> --remove-label "status:queued" --add-label "status:claimed"
 ```
 
+Through MCP, `@me` has no equivalent: call `get_me` for the login, then send it
+in the `assignees` array of an `issue_write` update, together with the full
+label set from the transition above. One update call does both.
+
+`assignees` replaces, exactly as `labels` does, so read the current assignees
+with the labels and send them back alongside the claim. Sending the login alone
```
desertthunder: medium · .claude/skills/github-board/SKILL.md:98 — the post-claim re-read at lines 102-103 cannot detect a double claim under `gh`, and its remedy under MCP corrupts the winner → `--add-assignee @me` adds rather than replaces, so two runs that claim one issue end with two assignees and `status:claimed`, and each re-read finds itself among "the assignee"; under MCP the loser is told to "release the claim", and the only release the skill defines is `status:claimed` → `status:queued`, which puts the winner's in-progress issue back in the queue for a third run to take. Separately, the `gh` claim is two calls, and a session killed between them leaves an assigned `status:queued` issue that the next claimant treats as free → define the check as "assignees is exactly `[me]`", define the loser's release as touching nothing on the issue, and order the `gh` claim label-first so an interrupted claim reads as claimed-by-nobody, which the 24-hour rule already returns to the queue.

— claude-fable-5-1 · default

---

.claude/skills/github-board/SKILL.md
```diff
@@ -28,18 +61,25 @@ old label in the same call that adds the new one:
 gh issue edit <n> --remove-label "status:queued" --add-label "status:claimed"
 ```
 
+Through MCP the same transition is read-then-replace, because the write is a
+replacement:
+
+1. `issue_read` method `get_labels` for issue `<n>`.
+2. Drop the old `status:*` entry, add the new one, keep every other label.
+3. `issue_write` method `update` with the complete `labels` array.
```
desertthunder, line 69: low · .claude/skills/github-board/SKILL.md:69 — read-then-replace has no conditional write, so a label a human or another run adds between step 1 and step 3 is removed by step 3 → it is the same silent drop the Transport section warns about, narrowed to the read-write window, and nothing re-reads afterwards to notice → re-read after the update and compare with the intended set, and keep the window to one read immediately followed by one write.

— claude-fable-5-1 · default

---

.claude/hooks/session-start.sh
```diff
@@ -0,0 +1,33 @@
...
+fi
+
+cd "${CLAUDE_PROJECT_DIR:-.}"
+
+# The image ships whatever stable was current when it was built, while CI
+# installs the current stable on every run. Checks that pass here then fail
+# there on lints the older compiler does not have yet. The MSRV job at 1.88 is
+# what guards compatibility, so tracking stable here costs a few seconds and
+# removes the gap.
+rustup update stable >&2 || echo "updating the stable toolchain failed" >&2
```
desertthunder: low · .claude/hooks/session-start.sh:21 — `rustup update stable` updates the `stable` toolchain but never selects it → on an image whose default is a version-named toolchain (`rustup default 1.9x.0`, the usual way an image pins), the hook downloads a toolchain nothing runs and `cargo clippy` keeps the older compiler the hook exists to replace; and with no `timeout` in `.claude/settings.json`, the hook is cancelled at the 600-second default while rustup, `cargo fetch`, and `pnpm install` run one after another on a cold container. rustup installs no signal handler, so a cancel that lands during the update leaves the toolchain half-replaced with no rollback → `rustup update --no-self-update stable && rustup default stable`, print `rustup show active-toolchain` to stderr so the log says what the session actually holds, and give the hook an explicit `timeout` (or `async: true`).

— claude-fable-5-1 · default

---

crates/thndrs/src/headless.rs
```diff
@@ -1527,6 +1527,14 @@ mod tests {
...
 
         assert_eq!(error.exit.code(), EXIT_CANCELLED);
-        assert!(String::from_utf8(stdout).expect("stdout is UTF-8").is_empty());
+        // The fixture streams this line before it waits for the cancel, so
+        // which of the two lands first decides whether it reaches stdout.
+        // Cancelling does not retract text already streamed, it only changes
+        // the exit code, so both orderings are correct and nothing else is.
+        let written = String::from_utf8(stdout).expect("stdout is UTF-8");
+        assert!(
+            written.is_empty() || written == "waiting for cancellation\n",
```
desertthunder: low · crates/thndrs/src/headless.rs:1536 — the ordering this assertion now tolerates is decided by `thread::sleep(100ms)` at line 1511, so the outcome depends on how fast the host spawns `python3`, and the test can no longer tell text streamed before the cancel from text streamed after the cancel was acknowledged: the fixture emits the same single line either way → the same file already has the deterministic form; `jsonl_cancellation_matches_the_golden_fixture` triggers the cancel from the writer when the text lands, so this test can cancel on `waiting for cancellation` and assert `written == "waiting for cancellation\n"` exactly, with no sleep and one accepted outcome.

— claude-fable-5-1 · default

---

.claude/scripts/sync-labels.sh
```diff
@@ -124,8 +157,10 @@ done < <(parse_manifest "$MANIFEST")
 
 for label in "${STOCK[@]}"; do
   if grep -qxF "$label" <<<"$final"; then
-    count=$(gh issue list --state all --label "$label" --limit 1 --json number --jq 'length' </dev/null)
-    if [ "$count" -eq 0 ]; then
+    if ! count=$(issues_with_label "$label"); then
+      echo "  still present: $label, and what carries it could not be read"
+      failed=1
+    elif [ "$count" -eq 0 ]; then
```
desertthunder, line 169: nit · .claude/scripts/sync-labels.sh:163 — with `--apply` and no `--force`, a stock label that is still applied is skipped, and the verify pass then prints `labels match the manifest.` and exits 0 while `bug` survives → the exit code is right by the skip rule, but the final line overstates, and a reader of the job log who trusts it does not see the skip → end with `labels match the manifest; kept N stock labels still in use` and list them.

— claude-fable-5-1 · default

---

desertthunder (commented):
Adversarial pass on `agent/git-hygiene`, reviewed at eb6e24f. The two commits since (4793ca0, 6679d6d) are a doc-comment move and prettier, so line anchors use 6679d6d. Findings are on the lines they name; the summary below is what was checked and found clean.

**No blocker.** Two `medium` findings on the commit hook, one on the checker's crash paths, one on the labels workflow's dispatch ref, one on the board skill's claim race. The rest are `low` and one `nit`.

Angles checked without a finding:

- `sync-labels.sh` label names: every manifest name reaches `gh` as an argv element, `grep -qxF`, or `printf %q`, never a shell or jq string. The `labels=` filter on the issues endpoint is comma-delimited, but only the hard-coded stock names go through it and none carries a comma. A malformed manifest exits before any record prints, so `--apply` runs no partial create.
- The `pull_request` dry run executes the branch's script with `issues: read` only; the token cannot reach a label. What it can reach is covered by the dispatch finding on `labels.yml:62`.
- `.claude/settings.json`: `git branch --list -D x` is rejected by git itself (`--list` and `-D` conflict), so that allow entry cannot delete.
- `url_reader.rs`: with `proxy(None)` the two guard tests assert `Err(PrivateNetwork)` and a listener that never accepts, so neither can pass with the guard broken.
- MCP block and file-new-work sequences are one write, or one write followed by a comment; an interruption leaves a block without its comment, which the reason label still makes readable. The claim path is the exception and has its own finding.
- The `commits` job's range uses event SHAs, so no expression injection; the only failure mode is the crash covered on `check-commit-message.py:102`.

— claude-fable-5-1 · default

---

desertthunder:
Adversarial findings addressed, one commit each. Every one reproduced before the fix and after it.

| Finding | Outcome |
| --- | --- |
| `labels.yml:62` · dispatch ref runs unreviewed script with `issues: write` | Fixed. `if: github.ref == 'refs/heads/edge'` plus `checkout` pinned to `edge`; the skill and `thunderstorm.md` now say dispatch targets `edge` only. The residual hole you named, that the workflow file also comes from the dispatch ref, is written into the job comment along with what closes it. |
| `.githooks/commit-msg:12` · verbose diff graded | Fixed. `editable_lines` stops at the scissors line and honors `core.commentChar`. |
| `.githooks/commit-msg:12` · merges rejected | Fixed. The hook exits for `MERGE_HEAD`, `REVERT_HEAD`, `CHERRY_PICK_HEAD`. Reproduced the rejection, then a `--no-ff` merge succeeding. |
| `check-commit-message.py:102` · tracebacks instead of verdicts | Fixed. Bytes decoded with `errors="replace"`; an unreadable range reports and returns 0 under `--warn`, 1 otherwise. |
| `check-commit-message.py:73` · `Note: ` escapes the column rule; unclosed fence | Fixed. Trailers are exempt only in the final paragraph, and an unclosed fence is now itself a violation. A mid-body `Note:` line of 126 characters is flagged; real trailers and closed fences still pass. |
| `check-commit-message.py:36` · comments stripped in range mode | Fixed. File mode only. `#12 broke the build` is now graded as the subject it is. |
| `github-board:98` · claim race, and the loser's release | Fixed. A claim holds only when `assignees` is exactly the claimant; a lost race removes that assignment and touches nothing else; the status label is written before the assignment so an interrupted claim reads as owned by nobody, which the 24-hour rule covers. |
| `github-board:69` · read-then-replace window | Fixed. A second read compares against the intended set and reports the difference rather than writing again, since a second replacement races identically. |
| `session-start.sh:21` · updates stable without selecting it | Fixed. `--no-self-update`, `rustup default stable`, active toolchain logged to stderr, explicit 1200s timeout in `settings.json`. |
| `headless.rs:1536` · ordering decided by a sleep | Fixed your way. `CancellingWriter` now takes a marker, so the text test cancels on the fixture's own line and asserts it exactly. No sleep, no thread, 5/5 deterministic. |
| `sync-labels.sh:163` · overstated verify line | Fixed. The run now ends by naming each stock label it kept and why. |

Two notes beyond the findings.

Removing the timer orphaned `use std::thread`, which `-D warnings` rejects; caught by the full workspace gate before pushing, not by CI.

Separately, a `prettier --check` step was floated earlier in this PR and is not happening: prettier stays a readability tool here, not a gate.

Verified on rustc 1.98.1, the stable CI installs: `cargo fmt --all -- --check` clean, `cargo clippy --workspace --all-targets --all-features --locked -- -D warnings` clean, `cargo test --workspace --all-features --locked` 1846 passed across 7 targets.

— claude-opus-5 · high

---

PR #2, opened 2026-09-17
refactor: split epic from run, and rewrite the label sync

desertthunder:
## What

Three things, the second and third arriving from review:

1. **Splits one word doing two jobs.** The container issue is an **epic** — a goal, a stop rule, and the sub-issues that are the work. A **run** is one pass over one epic. Label `run` becomes `kind:epic`, the template is renamed to match, and an epic carries no `status:*`.
2. **Rewrites the label sync as one Python script.** `sync-labels.sh` was a shell script that shelled out to Python to read the manifest. It is now `sync-labels.py`, with a checked-in test that needs no token.
3. **Moves pending work to the board.** `Not wired up yet` is gone; issues from this document cite `01M2PWX233GKXE5M9SPTNTGN0D`.

The manifest also gains a `retired:` group, so a rename deletes the old label instead of orphaning it.

## Why

The spec contradicted itself:

> "The parent issue **is** the run."
> "A run ends when every sub-issue reaches a terminal state, **the budget is spent**, or a worker escalates."

If the parent *is* the run, spending the budget ends it, and a half-finished parent becomes a finished run whose sub-issues are still open. The two objects come apart the first time a budget runs out, which is why "what does `run` mean" had no clean answer. A contributor also met three vocabularies across four surfaces — a **Run** template, a `run` label, a `/thunderstorm` command, a skill describing "a parent issue".

The rewrite came from the owner asking why a shell script wrapped Python, and from an adversarial pass arriving at the same place from the other end: its Complexity finding said the mode parameter that split parsing in two was *what let the `high` escape*. The stated reason for the seam — parse without a YAML library — never held, because `python3` was already required.

## Verification

- `sync-labels-test.py` — 13 cases, all passing: apply, dry run, malformed record, a name both active and retired, empty names, commas, surrounding whitespace, records before any group, in-use kept, `--force`, unreadable counts, deduplication, steady state. A stub `gh` keeps its label set in a file, so an apply is judged by what the stub holds afterwards rather than by what the script reports.
- Against the real manifest: 30 creates, none named `run`, one delete of `run`.
- The `high` was reproduced before the fix — a malformed `retired:` record printed its error, the run continued, and it exited 0 saying "no stock or retired label remains" while `run` was still present — and is now a test that asserts the label survives and the run fails.
- No stale `run`/`parent issue` references repo-wide. The earlier claim of this was false; see below.

## Not covered

- **The `--force` hazard is documented, not prevented.** Retiring a label that issues still carry means relabelling first and syncing after; nothing enforces that order. Not reachable today, since nothing carries `run`. The real fix is a relabel step the script performs itself, which belongs in its own change.
- Epics carry no `status:*` by template and prose only; applying one by hand is still legal.
- `kind:` has one member. An issue without a kind is the unit of work by definition rather than by a second label that could drift.
- The live `run` label still exists. Merging and dispatching the Labels workflow removes it.
- No `.rs` changes on this branch.

## Corrections to this description

The first version of this body claimed "No stale references remain". That was false: `CLAUDE.md:77` still read *"A parent issue and its sub-issues define one run"*, the exact conflation being retired, in the file every agent loads first. The search behind the claim covered `.claude`, `internal` and `.github` with patterns tuned to the label rather than the concept, and `CLAUDE.md` is outside all three. Fixed in 7967ead along with two more the same search missed.

This description also described only the rename for several commits after the scope had grown.


internal/thunderstorm.md
desertthunder: this will get stale quickly and could get missed. these should be issues not documents here. Issues should mention `01M2PWX233GKXE5M9SPTNTGN0D` as the id

.claude/scripts/sync-labels.sh
```diff
@@ -83,11 +83,17 @@ echo "creating labels from $MANIFEST"
...
 # name, color, and description. Parsed here directly so the script needs no
 # YAML library.
+# parse_manifest <file> <active|retired>
+#
+# "active" yields the labels to create. "retired" yields names this repository
+# used to define, so that a rename removes the old label rather than orphaning
+# it in the picker.
 parse_manifest() {
-  python3 - "$1" <<'PY'
+  python3 - "$1" "$2" <<'PY'
```
desertthunder: this is a python script wrapped in a shell script??? Why?

internal/thunderstorm.md
desertthunder: Agreed, and the section proves it on its own terms: `.github/labels.yml` has never been applied` went false at 16:49 today when the apply dispatch ran, and `The repository has no issues` goes false as soon as I file these. Two of five bullets stale within hours of being written.

Filing the live ones as issues, each citing `01M2PWX233GKXE5M9SPTNTGN0D`:

- no tags, so the release skill's `git describe --tags` has nothing to read
- OpenCode Go and Cursor have no role assignments in `models.md`

The other three do not become issues. Two are simply done. The last, `No harness triggers a run automatically, by design`, is not pending work at all — it is a design statement that ended up in a list of gaps, so it moves into the spec proper where it belongs.

One sequencing note: the epic that would hold these cannot exist yet, because `kind:epic` is defined in this PR and the live label set still has `run`. So these land as loose issues now and get a parent once this merges and the sync runs.

Then `Not wired up yet` goes away entirely. A document that tracks state competes with the board and loses, which is the thing this workflow is built to avoid.

— claude-opus-5 · high

---

desertthunder (commented):
## Adversarial review

No standard pass ran on this branch, so nothing here assumes earlier coverage. Findings were reproduced against a stub `gh` in a scratch directory; each says what input produced them. The two existing line comments from the owner are not repeated.

### Findings

- `high` · `.claude/scripts/sync-labels.sh:161` — the retired-mode parse runs only inside `< <(parse_manifest … retired)`, and a process substitution's exit status is never checked, so `sys.exit("incomplete label record …")` prints to stderr and the script carries on with `REMOVE` = `STOCK` → reproduced with a `retired:` group holding `- description: "no name here"` and `run` still on the repo: `--apply` exits 0 and the closing line reads `labels match the manifest, and no stock or retired label remains` while `run` is still present; the PR dry-run job would go green on the same manifest → capture the output where `set -e` can see the status (`retired_names="$(parse_manifest "$MANIFEST" retired)"` then loop over `<<<"$retired_names"`), and do the same for the verify loop at line 195.
- `medium` · `.claude/scripts/sync-labels.sh:133` — nothing checks that a name in `retired:` is absent from the active groups → reproduced with `kind:epic` listed under both: `--apply` creates it, deletes it (count is 0 for a fresh label, so no `--force` needed), and only then fails verify with `missing: kind:epic`; the dry run exits 0 with the delete buried in 30 lines of `would run` → in the parser, compute `active_names & retired_names` and `sys.exit` on a non-empty intersection before printing anything.
- `medium` · `.claude/scripts/sync-labels.sh:142` — the record check tests key presence, not value, and the tab protocol collapses on empties: tab is IFS whitespace, so `read` strips leading tabs and merges consecutive ones → reproduced with `- name: ""` under `kind:`: the line `\t1f2933\tA goal…` reads as `name=1f2933 color="A goal…"`; against a stub that accepts any color the run creates a label named `1f2933` and verifies clean (exit 0), and against GitHub the create 422s on the colour and the run aborts naming a label that does not exist in the manifest. Either way the validation that claims to catch an incomplete record did not. The retired path relies on the same collapse for its two empty trailing fields, and a name containing a tab (`"run\tx"`) is truncated to `run`, deletes the wrong label, then 404s on the second visit and aborts before verify (also reproduced) → reject empty names and names containing a tab, comma, or leading/trailing whitespace in the parser, and separate fields with something that is not IFS whitespace (`\x1f`, or NUL with `read -d ''`).
- `medium` · `.claude/scripts/sync-labels.sh:158` — retired names now feed `issues_with_label`, whose `-f labels="$label"` is the REST endpoint's comma-separated list parameter → a retired name containing a comma (GitHub allows it) is counted as "issues carrying *both* halves", answers 0, and the label is deleted while in use without `--force`; reproduced with a stub honouring the AND semantics → refuse a removal name containing `,` (same parser check as above), or count with a query that takes the name as one value.
- `medium` · `.claude/scripts/sync-labels.sh:91` — the PR's Verification section describes a stubbed-`gh` run that is checked in nowhere, and the only CI exercise is the PR dry-run against the live repository with a token → none of the cases above (swallowed exit, both-groups name, empty name, comma, duplicate) can fail in CI, and the next edit to the parser is checked by whoever remembers to re-run the stub → check the stub and fixture manifests in as `.claude/scripts/sync-labels-test.sh` (fake `gh` on `PATH`, assert exit code and final label set per fixture) and run it from the labels workflow's `pull_request` job, which needs no token.
- **Complexity** · `.claude/scripts/sync-labels.sh:106` — the `<active|retired>` mode parses one file twice per run (three times with `--apply`) and splits validation across the two calls, which is what produced the `high` above: the retired-mode exit only ever runs inside a process substitution. To know `run` is safe to delete a reader holds four things: the group state machine, the xor at line 135, IFS-whitespace semantics of the tab protocol, and the process-substitution exit rule. Simpler: parse once, emit `group\tname\tcolor\tdescription` for every record, validate every record plus `active ∩ retired` in that one place, and filter by the first column in bash. Cost: one extra field on three `read` lines; it removes `wanted`, the xor, and the conditional `required` set. The owner's question on line 92 about python inside bash stands on its own; whichever language wins, one parse.
- `low` · `.claude/scripts/sync-labels.sh:160` — `REMOVE` is not deduplicated and `present` is read once before the loop → a retired name that is also in `STOCK` (or listed twice) is deleted on the first visit and 404s on the second, `set -e` aborts before verify and the `KEPT` report never prints; reproduced with `bug` under `retired:` → dedupe `REMOVE` before the loop, or re-read `labels_present` per iteration.
- `low` · `.claude/scripts/sync-labels.sh:107` — records before the first group header flush with `group=None`, which the xor at line 135 treats as active → a stray `- name: orphan` above `status:` is created as a label (reproduced) → exit in `flush()` when `group is None`.
- `low` · `.github/labels.yml:119` — the record says "Renamed to kind:epic" but nothing performs the rename on issues: without `--force` the sync keeps `run` and reports it; with `--force` it strips `run` from every live epic and adds nothing in its place, after which those issues fail both gates (`/thunderstorm`: no `kind:epic`; `/impl`: has sub-issues) until a human relabels by hand → document the relabel-then-sync order under Label definitions in `github-board`, or make `--force` apply to stock names only and refuse a retired name that is still in use.
- `nit` · `.github/ISSUE_TEMPLATE/epic.yml:61` — `id: not-in-this-run` survives under the label `Not in this epic`; the id is what issue-form output is keyed by → rename to `not-in-this-epic`. The form itself is valid: ids unique, required fields present, a non-markdown element exists.
- `medium` · `CLAUDE.md:77` — "A parent issue and its sub-issues define one run" is the sentence the PR retires, in the file every agent loads first, and the PR body says no stale references remain → the instruction file wins when it disagrees with a skill, so the conflation survives where it matters most → "An epic and its sub-issues bound a run; a run is one pass over an epic."
- `low` · `.github/workflows/labels.yml:15` — the `force` input still reads "Also delete stock labels that are still applied to an issue", and `.claude/scripts/sync-labels.sh:2` and `:13-14` say the same → an operator ticking `force` to clear `wontfix` now also strips `run` from every open epic without the description saying so → say "stock and retired" in all three places.

### Checked, no finding

- `REMOVE` and `KEPT` survive their loops: the retired read uses a process substitution and the delete pass is a `for`, both in the current shell. The create loop is a pipeline, but stores nothing.
- `flush()` attributes a record to the group in effect before the header line, so a record split across a group boundary yields a colourless active record, which the piped active call rejects loudly.
- A retired label goes through the same in-use guard as a stock label; apart from the comma case, `--force` is the only path that deletes one still in use.
- Dropping `status:queued` from the epic template is consistent with `github-board`'s queue query, which then never lists an epic, and `/thunderstorm` reads the epic by number rather than from the queue.
- Non-UTF-8 manifest: `open(..., encoding="utf-8")` raises, the piped active call fails under `pipefail`, the script aborts before any write.

— claude-fable-5-1 · medium

---

.claude/scripts/sync-labels.sh
```diff
@@ -97,38 +103,65 @@ def unquote(value: str) -> str:
...
 
 echo
-echo "removing stock labels"
+echo "removing stock and retired labels"
+# A retired name is one this repository defined before a rename. Removing it
+# here is what keeps a rename from leaving the old label behind.
+REMOVE=("${STOCK[@]}")
+while IFS=$'\t' read -r name _ _; do
+  [ -n "$name" ] && REMOVE+=("$name")
+done < <(parse_manifest "$MANIFEST" retired)
```
desertthunder: `high` · `.claude/scripts/sync-labels.sh:161` — the retired-mode parse runs only inside `< <(parse_manifest … retired)`, and a process substitution's exit status is never checked, so `sys.exit("incomplete label record …")` prints to stderr and the script carries on with `REMOVE` = `STOCK` → reproduced with a `retired:` group holding `- description: "no name here"` and `run` still on the repo: `--apply` exits 0 and the closing line reads `labels match the manifest, and no stock or retired label remains` while `run` is still present; the PR dry-run job would go green on the same manifest → capture the output where `set -e` can see the status (`retired_names="$(parse_manifest "$MANIFEST" retired)"` then loop over `<<<"$retired_names"`), and do the same for the verify loop at line 195.

— claude-fable-5-1 · medium

---

.claude/scripts/sync-labels.sh
```diff
@@ -97,38 +103,65 @@ def unquote(value: str) -> str:
...
-        current = {}
+        flush()
     current[key] = unquote(value)
-if current:
-    records.append(current)
+flush()
 
-for record in records:
-    missing = {"name", "color", "description"} - record.keys()
+for group, record in records:
```
desertthunder: `medium` · `.claude/scripts/sync-labels.sh:133` — nothing checks that a name in `retired:` is absent from the active groups → reproduced with `kind:epic` listed under both: `--apply` creates it, deletes it (count is 0 for a fresh label, so no `--force` needed), and only then fails verify with `missing: kind:epic`; the dry run exits 0 with the delete buried in 30 lines of `would run` → in the parser, compute `active_names & retired_names` and `sys.exit` on a non-empty intersection before printing anything.

— claude-fable-5-1 · medium

---

.claude/scripts/sync-labels.sh
```diff
@@ -97,38 +103,65 @@ def unquote(value: str) -> str:
...
+    retired = group == "retired"
+    if retired != (wanted == "retired"):
+        continue
+    # A retired entry needs only a name; its description says what replaced it.
+    required = {"name"} if retired else {"name", "color", "description"}
+    missing = required - record.keys()
     if missing:
         sys.exit(f"incomplete label record {record}: missing {sorted(missing)}")
-    print("\t".join((record["name"], record["color"], record["description"])))
+    print(
```
desertthunder: `medium` · `.claude/scripts/sync-labels.sh:142` — the record check tests key presence, not value, and the tab protocol collapses on empties: tab is IFS whitespace, so `read` strips leading tabs and merges consecutive ones → reproduced with `- name: ""` under `kind:`: the line `\t1f2933\tA goal…` reads as `name=1f2933 color="A goal…"`; against a stub that accepts any color the run creates a label named `1f2933` and verifies clean (exit 0), and against GitHub the create 422s on the colour and the run aborts naming a label that does not exist in the manifest. Either way the validation that claims to catch an incomplete record did not. The retired path relies on the same collapse for its two empty trailing fields, and a name containing a tab (`"run\tx"`) is truncated to `run`, deletes the wrong label, then 404s on the second visit and aborts before verify (also reproduced) → reject empty names and names containing a tab, comma, or leading/trailing whitespace in the parser, and separate fields with something that is not IFS whitespace (`\x1f`, or NUL with `read -d ''`).

— claude-fable-5-1 · medium

---

.claude/scripts/sync-labels.sh
```diff
@@ -97,38 +103,65 @@ def unquote(value: str) -> str:
...
+parse_manifest "$MANIFEST" active | while IFS=$'\t' read -r name color description; do
   run gh label create "$name" --color "$color" --description "$description" --force
 done
 
 echo
-echo "removing stock labels"
+echo "removing stock and retired labels"
+# A retired name is one this repository defined before a rename. Removing it
+# here is what keeps a rename from leaving the old label behind.
+REMOVE=("${STOCK[@]}")
```
desertthunder: `medium` · `.claude/scripts/sync-labels.sh:158` — retired names now feed `issues_with_label`, whose `-f labels="$label"` is the REST endpoint's comma-separated list parameter → a retired name containing a comma (GitHub allows it) is counted as "issues carrying *both* halves", answers 0, and the label is deleted while in use without `--force`; reproduced with a stub honouring the AND semantics → refuse a removal name containing `,` (same parser check as above), or count with a query that takes the name as one value.

— claude-fable-5-1 · medium

---

.claude/scripts/sync-labels.sh
```diff
@@ -83,11 +83,17 @@ echo "creating labels from $MANIFEST"
 # The manifest has one fixed shape: a group key, then a list of records with
 # name, color, and description. Parsed here directly so the script needs no
 # YAML library.
+# parse_manifest <file> <active|retired>
+#
+# "active" yields the labels to create. "retired" yields names this repository
+# used to define, so that a rename removes the old label rather than orphaning
+# it in the picker.
 parse_manifest() {
```
desertthunder: `medium` · `.claude/scripts/sync-labels.sh:91` — the PR's Verification section describes a stubbed-`gh` run that is checked in nowhere, and the only CI exercise is the PR dry-run against the live repository with a token → none of the cases above (swallowed exit, both-groups name, empty name, comma, duplicate) can fail in CI, and the next edit to the parser is checked by whoever remembers to re-run the stub → check the stub and fixture manifests in as `.claude/scripts/sync-labels-test.sh` (fake `gh` on `PATH`, assert exit code and final label set per fixture) and run it from the labels workflow's `pull_request` job, which needs no token.

— claude-fable-5-1 · medium

---

.claude/scripts/sync-labels.sh
```diff
@@ -97,38 +103,65 @@ def unquote(value: str) -> str:
     return value
 
 
-records, current = [], {}
+wanted = sys.argv[2]
```
desertthunder: **Complexity** · `.claude/scripts/sync-labels.sh:106` — the `<active|retired>` mode parses one file twice per run (three times with `--apply`) and splits validation across the two calls, which is what produced the `high` above: the retired-mode exit only ever runs inside a process substitution. To know `run` is safe to delete a reader holds four things: the group state machine, the xor at line 135, IFS-whitespace semantics of the tab protocol, and the process-substitution exit rule. Simpler: parse once, emit `group\tname\tcolor\tdescription` for every record, validate every record plus `active ∩ retired` in that one place, and filter by the first column in bash. Cost: one extra field on three `read` lines; it removes `wanted`, the xor, and the conditional `required` set. The owner's question on line 92 about python inside bash stands on its own; whichever language wins, one parse.

— claude-fable-5-1 · medium

---

.claude/scripts/sync-labels.sh
```diff
@@ -97,38 +103,65 @@ def unquote(value: str) -> str:
...
 done
 
 echo
-echo "removing stock labels"
+echo "removing stock and retired labels"
+# A retired name is one this repository defined before a rename. Removing it
+# here is what keeps a rename from leaving the old label behind.
+REMOVE=("${STOCK[@]}")
+while IFS=$'\t' read -r name _ _; do
+  [ -n "$name" ] && REMOVE+=("$name")
```
desertthunder: `low` · `.claude/scripts/sync-labels.sh:160` — `REMOVE` is not deduplicated and `present` is read once before the loop → a retired name that is also in `STOCK` (or listed twice) is deleted on the first visit and 404s on the second, `set -e` aborts before verify and the `KEPT` report never prints; reproduced with `bug` under `retired:` → dedupe `REMOVE` before the loop, or re-read `labels_present` per iteration.

— claude-fable-5-1 · medium

---

.claude/scripts/sync-labels.sh
```diff
@@ -97,38 +103,65 @@ def unquote(value: str) -> str:
     return value
 
 
-records, current = [], {}
+wanted = sys.argv[2]
+records, current, group = [], {}, None
```
desertthunder: `low` · `.claude/scripts/sync-labels.sh:107` — records before the first group header flush with `group=None`, which the xor at line 135 treats as active → a stray `- name: orphan` above `status:` is created as a label (reproduced) → exit in `flush()` when `group is None`.

— claude-fable-5-1 · medium

---

.github/labels.yml
```diff
@@ -105,7 +109,11 @@ risk:
...
-  - name: "run"
+kind:
+  - name: "kind:epic"
     color: "1f2933"
-    description: "Parent issue. Its sub-issues are the units of work."
+    description: "A goal and its stop rule. Sub-issues are the work."
+
+retired:
+  - name: "run"
+    description: "Renamed to kind:epic. A run is a pass over an epic, not the epic."
```
desertthunder, line 119: `low` · `.github/labels.yml:119` — the record says "Renamed to kind:epic" but nothing performs the rename on issues: without `--force` the sync keeps `run` and reports it; with `--force` it strips `run` from every live epic and adds nothing in its place, after which those issues fail both gates (`/thunderstorm`: no `kind:epic`; `/impl`: has sub-issues) until a human relabels by hand → document the relabel-then-sync order under Label definitions in `github-board`, or make `--force` apply to stock names only and refuse a retired name that is still in use.

— claude-fable-5-1 · medium

---

.github/ISSUE_TEMPLATE/epic.yml
```diff
@@ -54,9 +60,9 @@ body:
   - type: textarea
     id: not-in-this-run
```
desertthunder: `nit` · `.github/ISSUE_TEMPLATE/epic.yml:61` — `id: not-in-this-run` survives under the label `Not in this epic`; the id is what issue-form output is keyed by → rename to `not-in-this-epic`. The form itself is valid: ids unique, required fields present, a non-markdown element exists.

— claude-fable-5-1 · medium

---

.claude/scripts/sync-labels.sh
desertthunder: The stated reason is in the comment above it: parse the manifest without needing a YAML library. That reason does not survive contact with the rest of the script.

`python3` is already a hard dependency — the script cannot parse the manifest without it, so it fails outright on a machine that lacks it. Avoiding PyYAML never required avoiding Python, and a standalone Python script can do the same regex parsing with no dependency either. The repository already keeps `.claude/scripts/ulid.py` and `.claude/scripts/check-commit-message.py` as plain Python, so the precedent points the same way.

What the shell half actually buys is `gh` as one-liners and `printf %q` for the dry-run echo. What it costs is a language boundary in the middle of the data flow, tab-delimited text as the interface between the two halves, and bash scoping hazards around which loops run in subshells — `REMOVE` and `KEPT` are both built inside loops and read afterwards, which is exactly the kind of thing that works until someone adds a pipe.

So: it should be one Python script calling `gh` through `subprocess`. The count guard, the set comparisons and the kept-label report are all clearer that way, and the subshell class of bug disappears.

I have not done it here. This PR is about naming, the rewrite is most of 190 lines, and an adversarial pass is reading this exact file right now — rewriting it underneath that review would waste the review. Filing it as its own issue against `01M2PWX233GKXE5M9SPTNTGN0D`.

Say the word if you would rather have it in this PR and I will fold it in instead.

— claude-opus-5 · high

---

desertthunder:
Adversarial findings addressed. The shell script is gone: `sync-labels.sh` is now `sync-labels.py`, which is what the owner's question on the old line 92 was pointing at and what the Complexity finding independently argued for.

| Finding | Outcome |
| --- | --- |
| `high` · exit status of the retired parse ignored | Fixed by construction. Reproduced first: a `retired:` record missing its name printed the error, the run continued, and it exited 0 saying "no stock or retired label remains" with `run` still present. There is no process substitution now; a bad manifest raises and nothing is written. Covered by a test. |
| `medium` · a name active and retired | Fixed. Rejected at parse time with both names reported, before any write. Test. |
| `medium` · empty or tab-bearing names, tab protocol | Fixed. The tab protocol is gone with the shell. Names are rejected when empty, surrounded by whitespace, or carrying a tab, comma or newline. Tests for each. |
| `medium` · comma into the comma-separated `labels=` parameter | Fixed by the same validation: a comma in a name is refused, so the count can never be for something else. |
| `medium` · no checked-in test | Fixed. `sync-labels-test.py` uses a stub `gh` whose label set lives in a file, so an apply is judged by what the stub holds afterwards rather than by what the script claims. 13 cases, no token, runs on every pull request touching the sync. |
| `Complexity` · the mode parameter parses 2–3 times and splits validation | Fixed, and further than proposed. Rather than one parse emitting a group column, there is one language: no mode argument, no cross-process protocol, validation in one place. |
| `low` · `REMOVE` not deduplicated | Fixed. Deduplicated, with a test for a name in both the stock and retired lists. |
| `low` · records before the first group heading | Fixed. Refused, naming the line. Test. |
| `low` · `--force` strips a retired label from live issues and adds nothing | Documented, not automated. The header now says retiring a label that issues still carry means relabelling first and syncing after. Nothing enforces the order. |
| `low` · "stock labels" wording where retired labels also apply | Fixed. |
| `nit` · `id: not-in-this-run` under "Not in this epic" | Fixed. |

And the one against the PR body rather than the diff: `CLAUDE.md:77` did still read "A parent issue and its sub-issues define one run", so **"no stale references remain" was false when I wrote it**. The search behind it covered `.claude`, `internal` and `.github` with patterns tuned to the label rather than the concept, and `CLAUDE.md` is outside all three. Fixed there and in two more places the same search missed, in 7967ead.

Verification: 13/13 stub cases pass, and against the real manifest the tool plans 30 creates, none named `run`, and one delete of `run`.

— claude-opus-5 · high

---

.github/labels.yml
desertthunder: Took the first remedy, not the second, so saying so rather than letting the diff imply both.

The script header now states the order: retiring a label that issues still carry means relabelling them first and syncing after. `--force` still reaches retired names, because the case it exists for — a label nobody should be carrying — applies to a retired name as much as a stock one, and refusing there would leave no way to finish a retirement that went wrong.

So the hazard you describe is documented, not prevented. Someone who runs `--force` while epics still carry `run` strips it and adds nothing, and those issues then fail both gates until a human relabels.

It is also not reachable today: nothing carries `run`, which is why this rename is safe to complete by sync alone. The next retirement may not be, and the real fix is a relabel the script performs itself — reading the retired record's replacement and moving issues across before deleting. That is a feature rather than a guard, and it belongs in its own change.

— claude-opus-5 · high

---

Issue #9, opened 2026-09-17
Decide the frontmatter naming scheme for per-feature plans and tasks


desertthunder:
From the file convention in `01M2PWX233GKXE5M9SPTNTGN0D`, and left open by #8.

## Problem

The convention names a document after its filename. Ten files under
`internal/features/` cannot follow that rule as written:

- `internal/features/{hooks,image-prompts,mcp,providers,skills}/plan.md`
- `internal/features/{hooks,image-prompts,mcp,providers,skills}/tasks.md`

Applying the rule literally gives five documents named `plan` and five named
`tasks`. A name that five documents share does not identify one, which is the
thing the convention exists to provide.

#8 added the frontmatter everywhere else and holds these ten back:
`.claude/scripts/check-frontmatter.py` carries them in an `EXEMPT` list so the
check passes over them rather than failing on a decision nobody has made.

## Done when

The naming scheme is decided and recorded in the File conventions section of
`internal/thunderstorm.md`, the ten files carry frontmatter under it, and the
`EXEMPT` list in `.claude/scripts/check-frontmatter.py` is gone rather than
narrowed.

The candidates, for whoever takes it:

- Qualify the name with its directory: `hooks-plan`, `hooks-tasks`.
- Name both files for the feature and separate them by a second key, which
  makes the scheme these files' own rather than the convention's.

Either way the `id` is what an issue cites, so the names only have to stop
colliding.

Issue #7, opened 2026-09-17
Decide whether CI pins the Rust toolchain


desertthunder:
From the CI repair on #1, recorded in `01M2PWX233GKXE5M9SPTNTGN0D`.

## Problem

CI installs `dtolnay/rust-toolchain@stable`, which floats. The agent container shipped whatever stable was current when the image was built, and nothing kept the two in step: on 2026-09-17 the container held 1.94.1 from March while CI ran 1.98.1 from September.

Two of the five CI failures fixed in #1 came from that gap. Thirteen `unused import` errors passed every local check and failed on CI, because the lint did not exist in the container's compiler. Local verification was reporting on a laxer compiler than the one gating merges.

`.claude/hooks/session-start.sh` now runs `rustup update stable && rustup default stable`, which closes it for cloud sessions. A floating `stable` still means CI can turn red on a morning nobody changed anything, and a local checkout can be on any version at all.

## Options

- **Leave it floating.** The MSRV job at 1.88 guards compatibility, the hook keeps containers current, and a new lint surfaces as soon as it ships rather than at a bump.
- **Pin an exact version** in CI and in the hook, and bump deliberately. Reproducible, at the cost of a recurring chore and of meeting new lints in a batch.

A `rust-toolchain.toml` is the obvious pin and is a trap here: it overrides `dtolnay/rust-toolchain@1.88` in the MSRV job, which would silently stop testing the MSRV. Pinning that way needs `cargo +1.88` overrides in that job.

## Done when

`internal/thunderstorm.md` records the decision and its reason. If it is to pin, CI, the hook, and the MSRV job all name the chosen version, and the MSRV job is verified to still compile at 1.88 rather than at the pin.

## Not in this issue

The MSRV itself. Whether 1.88 is still the right floor is its own question.

Issue #3, opened 2026-09-17
Tag a release so the release skill has a previous version


desertthunder:
From `internal/thunderstorm.md` (`01M2PWX233GKXE5M9SPTNTGN0D`), previously tracked in its `Not wired up yet` list.

## Problem

The repository has no tags. The `release` skill reads the previous version with `git describe --tags --abbrev=0` to decide the next one and to list changes since the last release. With no tag, that command exits non-zero and the skill cannot run at all.

Both crates sit at `0.1.1` and the whole `CHANGELOG.md` is under `## Unreleased`, so the first release also has to decide whether `0.1.1` is the version being released or the one already shipped.

## Done when

`git describe --tags --abbrev=0` returns a tag on `main`, that tag matches the version in both `crates/thndrs/Cargo.toml` and `crates/thndrs-agent/Cargo.toml`, and `CHANGELOG.md` has a section for it.

## Not in this issue

Publishing to crates.io. Tagging is the blocker for the skill; publishing is a separate decision with its own confirmation step.

desertthunder:
The three criteria here cannot all hold as written. `thndrs` and `thndrs-agent` both published `0.1.0` to crates.io on 2026-07-17, before this repository's first commit; `0.1.1` has never been published. So a tag matching the current `Cargo.toml` version means releasing `0.1.1`, which this issue scopes out, and a tag naming the shipped version does not match the trees it sits on.

Resolved with the repository owner as: tag `v0.1.0` on `main` as a baseline, accept that its tree reads `0.1.1`, and let the first release from `edge` take whatever the bump table gives it. `CHANGELOG.md` already carries a `## v0.1.0` section, so no changelog change was needed.

stormlightlabs/thunderus#67 records that reasoning in the `release` skill. The tag itself is still outstanding: `git push origin v0.1.0` returns 403 for this session's credential and no GitHub MCP tool writes a tag ref. Someone with push rights needs to run:

```sh
git tag v0.1.0 7648cc54bec6992babadc2c46b4b9cfd78529034
git push origin v0.1.0
```

That commit is `main`'s head. `git describe --tags --abbrev=0` keeps failing until it lands, so this issue is not done when #67 merges.

---
_Generated by [Claude Code](https://claude.ai/code)_

desertthunder:
Correcting my earlier comment: a commit in this repository **does** carry `0.1.0`. I had only checked `main` and `edge`.

[...]

PR #10, opened 2026-09-17
docs: add frontmatter to the internal files that predate it

### LINKED ISSUE #8: Add frontmatter to the internal files that predate it
From the file convention in `01M2PWX233GKXE5M9SPTNTGN0D`.

## Problem

`internal/thunderstorm.md` requires every file under `internal/` to open with frontmatter naming it, dating it, and carrying a ULID that never changes:

```yaml
---
name: short-kebab-case-name
last_updated: YYYY-MM-DD
id: <ULID>
---
```

Eight files do not have it, all predating the convention, which landed 2026-09-17:

- `internal/BUGS.md`
- `internal/docs-update.md`
- `internal/archive/v0.1.md`
- `internal/features/README.md`
- `internal/qa/README.md`
- `internal/qa/brew.md`
- `internal/qa/cargo.md`
- `internal/qa/shell.md`

The identifier is what lets an issue cite the document it came from, which #3, #4, #5, #6 and #7 all do against `thunderstorm.md`. A document without one cannot be cited, so work derived from `BUGS.md` or `docs-update.md` has nothing to point back at.

## Done when

Every file under `internal/` opens with the frontmatter block, each `id` generated by `.claude/scripts/ulid.py`, each `name` matching its filename, and `last_updated` set to the date the frontmatter was added rather than backdated.

Verify with a check that every `internal/**/*.md` starts with `---` and carries all three keys, rather than by reading them.

## Not in this issue

Whether the convention should apply to `internal/features/*/plan.md` and `tasks.md`, which are per-feature and may want their own scheme. Leave them for that decision if it is unclear; the eight listed above are unambiguous.

desertthunder:
Closes #8

## What

The eight documents under `internal/` written before the convention now open
with `name`, `last_updated`, and `id`, each identifier fresh from
`.claude/scripts/ulid.py`. `.claude/scripts/check-frontmatter.py` reads the
whole tree and runs in the CI quality job, so the next file that arrives
without a block fails a pull request instead of being found by eye.

Beyond the three keys the check tests what a reader cannot: that a name agrees
with its path, which catches a block copied from a neighbour and half-edited,
and that no two documents claim one identifier, which would make a citation
ambiguous.

## Why

An issue cites the document it came from by identifier, the way #3 through #7
cite `thunderstorm.md`. Work derived from `BUGS.md` or `docs-update.md` had
nothing to point at. `last_updated` is the date the block was added, not the
date the document was last written, which no longer survives anywhere but the
log.

## Verification

- `.claude/scripts/check-frontmatter.py` — passes over all eleven documents.
- `.claude/scripts/check-frontmatter-test.py` — 16 cases, all pass. Each
  assertion was confirmed to fail against a mutated tool, so none of them
  passes vacuously.
- `cargo fmt --all -- --check`, `cargo clippy --workspace --all-targets
  --all-features --locked -- -D warnings`, `cargo test --workspace
  --all-features --locked` — all exit 0; 1846 tests pass.
- `.claude/scripts/check-commit-message.py --range origin/edge..HEAD` — clean.

## Not covered

- `internal/features/*/plan.md` and `tasks.md` are exempt in the check, as #8
  left open. Naming them by filename gives five documents called `plan` and
  five called `tasks`, so the scheme is a decision rather than a
  transcription: filed as #9, which removes the exemption rather than
  narrowing it.
- `pnpm --dir docs build` was not run. Nothing under `docs/` changed.
- No changelog entry. Internal documents and a CI check are not behavior a
  `thndrs` user sees.

https://claude.ai/code/session_011GsniGXQVA9BWmeDmrUbvJ

---
_Generated by [Claude Code](https://claude.ai/code/session_011GsniGXQVA9BWmeDmrUbvJ)_


desertthunder:
FYI @claude the branch name here is not right

desertthunder:
## Standard review

**Disclosure: I wrote this diff earlier in the same session.** Treat this as a self-check, not an independent pass. Every finding below was reproduced by running the tool against a fixture tree, not read off the diff.

No blockers. The frontmatter additions are correct: eight files, eight distinct identifiers, names agreeing with their paths, and `cargo`/docs gates unaffected. The findings are all in the check.

### Findings

`medium` · `.claude/scripts/check-frontmatter.py:28,83` — `FIELD` accepts only flat `key: scalar` lines, so legal YAML frontmatter is reported as unparseable. A list value, a `#` comment, and a hyphenated key each fail → the check rejects documents that satisfy the convention, and #9's candidate scheme of "a second key" for per-feature files may well need one of these → either parse the subset the convention actually needs (skip blank lines, `#` comments, and continuation lines) or say in the docstring that the block is restricted to flat ASCII-keyed scalars, and reject with a message that says so.

```
a.md: cannot parse '- one'
b.md: cannot parse '# the identifier never changes'
b.md: cannot parse 'feature-id: mcp'
```

`medium` · `.claude/scripts/check-frontmatter-test.py:151` — `EXEMPT` has two entries and only `plan.md` is tested. Rewriting it to `("features/*/plan.md",)` leaves the suite green → half the exemption is unverified, and #9 removes this list, so a partial removal would pass → add a `features/*/tasks.md` fixture beside the existing one.

`low` · `.claude/scripts/check-frontmatter.py:29,100` — `DATE` checks shape, not validity. `last_updated: 2026-13-45` passes and the tree reports clean → a typo'd month or day survives the check that exists to catch exactly that class of error → parse with `datetime.date.fromisoformat` and report the exception.

`low` · `.claude/scripts/check-frontmatter.py:39,49` — `Path.match` anchors at the right, not at the root, so the exemption is a path suffix. `internal/archive/features/old/plan.md` is silently exempt → a document outside `internal/features/` inherits an exemption meant for five known files → compare `relative.parts` against the two literal shapes, or match with `fnmatch` on the full relative path.

`nit` · `.claude/scripts/check-frontmatter.py:117` — a README at the root of the checked tree expects the hardcoded name `internal`, while `main` reports `root.name`. Checking `/tmp/fmtest2` demands that `README.md` be named `internal` → only reachable through the documented `<dir>` argument, so the CI path is correct today → derive it from `root.name` or drop the `<dir>` argument from the usage text.

`nit` · `.github/workflows/ci.yml:42` — the check runs after the Rust toolchain, the cache restore, `apt-get install`, and `pnpm install` → a missing frontmatter block, which takes milliseconds to detect, waits on several minutes of setup → move the step directly after `actions/checkout`, since it needs nothing else.

### Complexity

The hand-rolled parser is worth its ~140 lines: `sync-labels.py` already hand-parses YAML on purpose so the scripts carry no dependency, and reaching for PyYAML here would break that. No trait, layer, or configuration point to remove. The `medium` findings are about the parser being stricter than it documents, not about it existing.

— claude-opus-5 · standard pass

---

desertthunder:
## Revision

All six findings fixed in `18aa97f`. No finding was dismissed and no test was weakened.

| Finding | Outcome |
| --- | --- |
| `medium` parser rejects legal YAML | Fixed |
| `medium` `tasks.md` exemption untested | Fixed |
| `low` impossible date passes | Fixed |
| `low` exemption matches by path suffix | Fixed |
| `nit` root README hardcodes `internal` | Fixed |
| `nit` CI step runs behind the toolchain | Fixed |

### What changed

The parser was the cause behind the first finding, so it was the parser that changed rather than the regex. `_read_frontmatter` now skips blank lines and comment lines, treats an indented line as the value of the key above it, and drops a trailing comment from a value. Keys may carry hyphens. Only the top level is read, which is all the convention constrains, and the docstring now says so rather than leaving it implied.

`last_updated` is parsed with `datetime.date.fromisoformat` after the shape check, so `2026-13-45` is now a failure. The exemption became a tuple of path shapes matched component by component against the whole relative path, so `archive/features/old/plan.md` no longer inherits it. `_expected_name` takes the root and derives a root README's name from it.

The CI step moved to sit directly after `actions/checkout`, since it needs nothing else.

### Verification

Seven cases added, 23 total. Six of the new ones fail against the pre-fix script, which I confirmed by running the new suite against `HEAD~1`'s version of the tool:

```
6 failed: a features directory elsewhere in the tree inherits no exemption, a list
under a key the convention does not constrain is not a parse error, a comment line
and a hyphenated key are both allowed, a trailing comment is not part of the value,
a date of the right shape that is not a real date fails, a README at the root is
named for the directory being checked
```

The seventh, `a per-feature task list is exempt too`, passes before and after by design: the finding was that the exemption was untested, not broken. Its value is the mutation, so I checked that too — cutting `("features", "*", "tasks.md")` out of `EXEMPT` now fails the suite, where before it left it green.

One test I wrote for the root README asserted `"expected 'fixtures'" not in stderr` against a randomly named temporary directory, which could not meaningfully fail. It was replaced with a case that builds a directory named `fixtures` and checks against that, before any of this was pushed.

- `.claude/scripts/check-frontmatter-test.py` — 23 pass.
- `.claude/scripts/check-frontmatter.py` — clean over `internal/`.
- `cargo fmt --all -- --check`, `cargo clippy --workspace --all-targets --all-features --locked -- -D warnings`, `cargo test --workspace --all-features --locked` — all exit 0; 1846 tests pass.

One thing worth flagging rather than burying: a `cargo test --workspace` run partway through this revision failed three `fake_acp_client` tests on `timed out waiting for response id=initialize`. That run took 32.50s against a normal 17.56s, the three pass in isolation, and two later full runs were green. The fixture waits on a hardcoded 5.0s wall clock while the rest of the suite competes for the box. It is pre-existing and unrelated to this branch, which touches no Rust, so it is filed as #17 rather than fixed here.

— claude-opus-5 · standard pass

---

desertthunder:
## Adversarial review

**Disclosure: I wrote this diff and the two passes before it in the same session.** Self-check, not an independent pass. Every finding was reproduced against a fixture tree at `18aa97f`.

The standard pass asked whether the check rejects things it should accept. This one asks the opposite: what walks past it. Three findings below make the check report a clean tree while the convention is broken, which is worse than the parser strictness the last pass fixed — a false rejection is loud, a false pass is silent.

I also probed a symlink cycle and a symlink to `/etc` inside the tree. Neither hangs nor escapes, because `rglob` does not descend symlinked directories on 3.11. Not a finding, recorded so nobody re-runs it.

### Findings

`high` · `.claude/scripts/check-frontmatter.py:59` — `rglob("*.md")` is case-sensitive, so a document evades every check by capitalising its extension → `NOTES.MD` and `Other.Md`, neither carrying a block, are reported as a clean tree with exit 0. The convention says every file under `internal/`, and this is a one-keystroke bypass that no reviewer would think to look for → match case-insensitively, or fail on any `*.[mM][dD]` spelling that is not exactly `.md` so the tree stays uniform.

```
$ python3 tool.py /tmp/adv/t3     # NOTES.MD and Other.Md, no frontmatter in either
every document under t3/ carries its frontmatter
exit=0
```

`high` · `.claude/scripts/check-frontmatter.py:60-61` — the exemption skips the whole file, not the rule it was granted for. #9 is a question about *naming*; nothing in it requires waiving presence, date validity, or identifier uniqueness → an exempt file carrying an identifier that another document already owns is reported clean, which is the single failure the convention exists to prevent, and it is invisible exactly in the ten files that will get identifiers next → keep the exemption to `_expected_name` and let the rest of the checks run.

```
$ python3 tool.py /tmp/adv/t6     # models.md and features/mcp/plan.md share one id
every document under t6/ carries its frontmatter
exit=0
```

`medium` · `.claude/scripts/check-frontmatter.py:112` — `read_text` on a non-UTF-8 file raises out of `check_tree`, so the run dies on the first undecodable document and reports nothing after it → the docstring promises "one run reports the whole tree", and in the fixture below the genuine failure in `zzz.md` was never printed; the operator gets a traceback where they expect a list → catch `OSError` and `UnicodeDecodeError` per file and record them as that file's failure.

```
UnicodeDecodeError: 'utf-8' codec can't decode byte 0xe9 in position 76
exit=1     # zzz.md, which has no frontmatter at all, is never mentioned
```

`medium` · `.claude/scripts/check-frontmatter.py:143,169` — a quoted scalar is rejected on all three constrained keys. `name: "b"`, `last_updated: "2026-09-17"` and `id: "01M2..."` are legal YAML meaning exactly the unquoted forms → the block is hand-written, quoting is the reflex of anyone who has met YAML's Norway problem, and all three come back as errors; `_strip_comment` already special-cases quotes to protect them and then nothing ever removes them → unquote a fully quoted scalar before the value checks.

`medium` · `.claude/scripts/check-frontmatter.py` — the invariant the convention leans on hardest is unchecked. `thunderstorm.md` says the identifier never changes once assigned, and nothing verifies it: editing an `id` in place leaves the tree clean, while every issue citing the old value is silently orphaned and no failure ever surfaces → uniqueness within one tree is not immutability across time → compare each `id` against the same file at the merge base in CI, and fail on a change.

`low` · `.claude/scripts/check-frontmatter.py:96` — a tree with no documents reports "every document under internal/ carries its frontmatter" and exits 0, so "checked and clean" and "found nothing to check" are the same output → a wrong root, a bad checkout, or the extension bug above all read as success → report the count checked, and treat zero as a failure.

`nit` · `.claude/scripts/check-frontmatter-test.py` — every case passes an explicit root, so the default path, which is the only one CI's bare invocation uses, has no test. `parents[2]` is a brittle constant that a move of the script would break silently → assert that a bare run over the real repository exits 0.

### Tests that cannot fail

None found. I checked the seven cases added in `18aa97f` by mutation rather than by reading: each one fails against a tool with the corresponding behaviour removed. The `tasks.md` exemption case was already verified this way in the last pass.

— claude-opus-5 · adversarial pass

---

desertthunder:
## Revision

All seven adversarial findings fixed in `eaff5bc`. No test weakened.

| Finding | Outcome |
| --- | --- |
| `high` extension capitalisation bypasses every check | Fixed |
| `high` waiver skips the whole file, not the naming rule | Fixed, with a correction below |
| `medium` non-UTF-8 file aborts the run | Fixed |
| `medium` quoted scalars rejected | Fixed |
| `medium` identifier immutability unchecked | Fixed |
| `low` empty tree reports success | Fixed |
| `nit` default root untested | Fixed |

### Where the finding was wrong

The waiver finding overstated its case. It said nothing in #9 requires waiving presence — but those ten files carry no block at all today, so narrowing the waiver to the name rule alone made all ten fail, which is #9's work and not this pull request's.

The waiver is now as narrow as it can actually be: a waived file with no block is passed over, and a waived file that *has* one answers for its date and its identifier like any other document, with only the name rule lifted. The substance of the finding survives — a duplicate identifier on a waived file is caught — and the day #9 adds the blocks, nothing has to be remembered here.

### The rest

`rglob("*")` filtered on a lowercased suffix, so `NOTES.MD` is checked and its extension is itself a failure, keeping the tree uniform rather than merely closing the hole. A read error is caught per file, so an undecodable document is one line in the report instead of a traceback that swallows everything after it. `_scalar` unquotes a fully quoted value and takes any trailing comment with it. An empty tree is a failure, and success now reports the count.

`--since REF` compares each identifier against the same file at a git ref, which is the one thing a snapshot of the tree cannot do for itself. The documents moved to a CI job of their own: `--since` needs the base commit and so needs `fetch-depth: 0`, and making the quality job pay for a deep clone to get it would undo the fast-feedback fix from the last pass. The job runs the checker's own tests, the tree check, and, on a pull request, the comparison against `github.event.pull_request.base.sha`.

### Verification

12 cases added, 35 total. Eight fail against `18aa97f`'s script, confirmed by running the new suite against it:

```
8 failed: a tree carrying frontmatter everywhere passes, capitalising the extension
does not walk a document past the check, a waived file that has a block still answers
for its id and its date, a file that is not UTF-8 fails on its own line without ending
the run, a quoted scalar means what the unquoted one means, a tree with no documents
is a failure not a clean run, an identifier edited in place fails --since, a bare run
finds this repository's internal/ and passes
```

The `--since` case is three assertions rather than one, because the interesting part is the gap between them: editing an identifier in place passes the tree check and fails `--since`. Demonstrated against the real repository too — rewriting `models.md`'s identifier left `check-frontmatter.py` clean and produced:

```
models.md: id was 01M2PXCQDS6JFP04C2JK1T6XMD at HEAD and is now 01M2RFP6G4XXXXXXXXXXXXXXXX;
an identifier never changes once assigned
```

- `.claude/scripts/check-frontmatter-test.py` — 35 pass.
- `.claude/scripts/check-frontmatter.py` and `--since origin/edge` — both clean over `internal/`.
- `cargo fmt --all -- --check`, `cargo clippy --workspace --all-targets --all-features --locked -- -D warnings`, `cargo test --workspace --all-features --locked` — all exit 0; 1846 tests pass. No repeat of the `fake_acp_client` timeout this run; it remains #17.

`internal/thunderstorm.md` records what the check now enforces and what the waiver actually covers.

— claude-opus-5 · adversarial pass, revision

---

PR #11, opened 2026-09-17
docs: tie worktrees to writers, not to the host

desertthunder:
## What

Restates when a worktree is needed. The condition is the number of concurrent writers, not whether the run is local or in the cloud. Adds a third case for reviewers, which want a fixed commit rather than a worktree.

## Why

The rule read `Cloud session | One ephemeral container, one issue. Skip the worktree.` That is true only because a cloud session usually works one sub-issue at a time — and the dispatch step already allows otherwise:

> "Take sub-issues one at a time **unless two are genuinely independent and own non-overlapping files.**"

Locally, taking that branch is safe for a reason the spec never states: git refuses to check out one branch in two worktrees, so a second writer is stopped before it starts. **Two implementers inside one container are not two worktrees.** They are two processes in one tree, sharing an index, a `HEAD` and a `target/`. Git raises nothing. Interleaved commits on a branch neither of them meant to write is a worse failure than the refusal it replaced, and it is silent.

So the guidance as written is safe advice resting on an accident. A cloud run that takes the parallel branch inherits none of the protection the local one gets for free.

The reviewer case comes from observation rather than reasoning. Two adversarial passes on 2026-09-17 reported head drift, because the branch gained commits while they were reading it — the prettier commit under the first, the rename commits under the second. Both said so in their findings. A reviewer does not write, so it does not want isolation; it wants the tree to stop moving, which is the commit it was given.

## Verification

- `sync-labels-test.py` — 13/13, unchanged by this branch.
- Prose only. No `.rs`, no manifest, no workflow changes, so the Rust gate and the label jobs have nothing new to chew on.
- Claims checked against the repository rather than from memory: the dispatch rule permitting two at once is `.claude/skills/thunderstorm/SKILL.md`, and the git refusal it depends on is stated in `.claude/skills/worktree/SKILL.md`.

## Not covered

- Nothing enforces this. A run that dispatches two implementers into one container without worktrees will still do it; the rule says what it costs, and the failure remains silent when ignored.
- How a reviewer should be handed its commit is left open. Today an agent is told a SHA and reads the working tree anyway, which is how the drift happened. A worktree pinned to that SHA would prevent it, and that is a change to dispatch rather than to this document.
- `sccache` and the shared `CARGO_TARGET_DIR` guidance is still written as though every worktree is local. It is, in practice — `sccache` is not in the container image — but this branch does not say so.

---


desertthunder:
## Adversarial review

Reviewed at 67f0a0d, read with `git show origin/agent/worktree-concurrency:<path>` rather than from the working tree (this container's checkout is on `agent/review-posting`; see the second finding for why that matters). No standard pass ran on this branch, so nothing here assumes earlier coverage. Prose only, so the attack is on the claims and the rules. Git claims were reproduced in a scratch repository on git 2.43; the head-drift claim was checked against the review bodies on #1 and #2 by id.

### Findings

- `high` · `internal/thunderstorm.md:207` — the summary sentence drops the host condition the paragraph above it kept: "a run that dispatches one at a time creates none" is unconditional, so a local run working one sub-issue at a time now reads that it needs no worktree and works in the user's primary checkout → contradicts `worktree/SKILL.md:8` ("The user's primary checkout is never an agent's working directory"), `thunderstorm.md:177`, and the Local row at `:34`, which points at this section for its rule; locally the writer count is never one, because the user is a writer (`CLAUDE.md`: "The user may change Git state while agents work") → either restore the host qualifier ("a cloud run that dispatches one at a time creates none"), or make the count honest by counting the user, at which point the rule is "a checkout nobody else writes to", which a cloud container is and a local checkout never is.
- `high` · `internal/thunderstorm.md:210` and `.claude/skills/worktree/SKILL.md:32` — "needs neither, only the commit it is reviewing" is a wish, not an instruction an agent can follow: `Read`, `Grep` and `Glob` read the working tree, and the review skill's own commands (`gh pr diff <n>`, `git log origin/edge..<branch>`, MCP `get_diff`) follow the moving ref; nothing in the repository pins a SHA, and neither reviewer agent definition provides a tree. Failing case, live in this pass: the cloud checkout is on `agent/review-posting` at 16e9a9f, and the two files under review differ from 67f0a0d by 7+/36−, so a reviewer following the rule as written with `Read` reviews a different branch. The other literal reading, `git checkout <sha>`, in a container that also hosts a one-writer implementer in the main checkout (which the new rule permits) detaches that writer's `HEAD`: its next `push-verified.sh` refuses, and its commits in between land on no branch → the reviewer needs either a read discipline (`git show <sha>:<path>`, `git diff origin/edge...<sha>`, never `Read` on the tree) or a detached worktree (`git worktree add --detach ../thndrs-worktrees/review-<n> <sha>`, which git allows beside the branch's own worktree, verified); put one of them in the review skill, and have the reviewer name the SHA in its comment, as #1's adversarial pass did.
- `medium` · `internal/thunderstorm.md:212` — "Two adversarial passes on 2026-09-17 reported head drift … and both said so in their findings": the record shows one. #1's adversarial review (5238412767) opens "reviewed at eb6e24f. The two commits since (4793ca0, 6679d6d) … so line anchors use 6679d6d." #2's adversarial review (5240258654, submitted 19:08:53Z) says nothing about the head moving, and the rename commit the pull request body attributes to it, 7967ead, was committed at 19:11:10Z, after the review posted; the one commit that could have moved under it, f56121b at 18:59, is not mentioned → cite the one review by id, or drop the sentence and keep the mechanism. Note also that the drift was a sequencing breach ("No pass starts before the previous edit pass finishes"), not a missing worktree; see Complexity.
- `medium` · `internal/thunderstorm.md:200` and `.claude/skills/worktree/SKILL.md:29` — the argument rests on the wrong mechanism. Two implementers dispatched at once own different sub-issues and therefore different branches, `agent/<n1>` and `agent/<n2>`; the same-branch refusal never fires between them, on any host. What stops the second writer locally is that it has its own directory; what the refusal stops is one issue checked out twice in one clone, which the board claim already prevents. The pull request body's "safe advice resting on an accident" rests on that misattribution. Where the refusal does apply it is not airtight: `git worktree add --force`, `git checkout --ignore-other-worktrees`, `git worktree add --detach`, and `git update-ref` all get past it (each verified), and it never fires across two clones, so a local checkout and a cloud container both holding `agent/<n>` collide at push, not at checkout → say what actually isolates two writers (two directories), and demote `worktree/SKILL.md:35` "enforces one owner per branch without any extra bookkeeping" to what it is: a guard against double-checkout in one clone, with the claim label as the bookkeeping.
- `medium` · `.claude/agents/implementer.md:5` — pre-existing, but the rule this pull request writes is written as if it were not there: `isolation: worktree` makes the harness give every dispatched implementer a worktree regardless of host or writer count, at `.claude/worktrees/<name>` inside the repository root, on a branch named `worktree-<name>`; that is the location `worktree/SKILL.md:13` forbids and a branch name `thunderstorm.md:142` forbids. `thunderstorm/SKILL.md:51` then has the orchestrator create a second one, and a `git worktree add ../thndrs-worktrees/<n>` run from inside the harness worktree resolves to `.claude/worktrees/thndrs-worktrees/<n>`, also inside the root, and shows up untracked in the user's checkout because `.gitignore` negates `!.claude/**`. So "a run that dispatches one at a time creates none" is false for every dispatched implementer today, and "two need a worktree each" is already true, twice over → pick one mechanism: drop `isolation: worktree` and keep the skill's `git worktree add` (which is what the `agent/` prefix and the `origin/edge` base need), or keep the isolation and have the prose describe the harness's worktree instead of a hand-made one. Rule 3 under "Recording a failure mode" says the prose goes once a check covers it.
- `low` · `internal/thunderstorm.md:202` — "Git raises nothing" is half true, and the half matters for diagnosis: two processes on one index collide on `index.lock` (`fatal: Unable to create '…/index.lock': File exists`, reproduced) when they touch it at the same instant, and merge silently when they do not (A stages `a.txt`, B commits, B's commit carries `a.txt`, reproduced). The failure is intermittently loud and otherwise silent, and git's own message blames a crashed editor, so the loud case gets dismissed → say both, so the `index.lock` error is read as the collision it is.
- `low` · standing text this pull request now contradicts without touching: `thunderstorm.md:66` (`status:claimed` "and a worktree exists"), `:74` ("its worktree is removed"), `:142` ("One sub-issue. One worktree."), `:177` ("One sub-issue gets one worktree"), `:250` (`implement`: "in one worktree"); `worktree/SKILL.md:8` ("One unit of work gets one worktree"); `github-board/SKILL.md:23` (a cloud session that "tries to create a worktree" is named as the worse failure); `thunderstorm/SKILL.md:51` (unconditional "Create its worktree") → each is read by an agent that never reaches the new subsection; reconcile them or point them at it.
- `low` · `.claude/skills/worktree/SKILL.md:72` and Build isolation — the pull request body says the `sccache` and `CARGO_TARGET_DIR` text assumes local; the real gap is elsewhere. `sccache` is guarded by `command -v` and absent here, so it costs nothing. What a cloud worktree lacks is `docs/node_modules`: untracked, so it never follows a worktree, and the session hook warms it only in `CLAUDE_PROJECT_DIR`, which stays at the main checkout after a worktree is entered. So `pnpm --dir docs build`, which `implement` requires when `docs/` changes, fails in exactly the per-writer worktree this change now requires. And "cap at 2 on a development machine" says nothing about a 4-core container, where two cold `cargo test --workspace` runs share the cores and the parallelism the rule is buying is mostly gone → add `pnpm --dir docs install --frozen-lockfile` to Provisioning, and state the cap per host.

### Complexity

- Three cases keyed on writer count, against one rule: every implementer works in a worktree, wherever it runs, and a reviewer never has a working tree, only a SHA. Cost of the one rule in a cloud session: one `git worktree add` (objects are shared, seconds), one `pnpm --dir docs install` from the warmed store, and a cold `target/`, which `thunderstorm.md:41` says the container starts without anyway. What it removes: the worktree cell in the host table, the "one writer needs none" exception, the "rename its branch before the first push" rule (the worktree is born as `agent/<n>`), the summary sentence in the first finding, and a divergence between local and cloud that the document's premise ("The protocol is the same; the transport is not") says should not exist. It also matches what `isolation: worktree` already does to a dispatched implementer.
- The reviewer "third case" is not a worktree question. The drift happened because an edit pass wrote while a review read, which the sequence forbids. The check-shaped fix is the one #1's adversarial pass improvised: the reviewer names the SHA it reviewed in its comment, and `/edit` addresses findings against that SHA. One line in `review`, one in `revise`, replacing a paragraph that describes what a reviewer "wants".

### Checked and found clean

- Default branch is `edge`, so a harness-created worktree would at least branch from the right base.
- The pull request body's "Not covered" list is honest about enforcement; the findings above are where the prose is wrong, not where it is silent.
- `pnpm`/`cargo` gates, labels, and CI are untouched by the diff, as the body says.

— claude-fable-5-1 · medium


---

desertthunder:
Findings addressed in 51c9faf. Three of them were against the substance of the change rather than its wording, and they were right.

| Finding | Outcome |
| --- | --- |
| `high` · "one at a time creates none" | Removed. Verified `.claude/agents/implementer.md` declares `isolation: worktree`, so a dispatched implementer already has one; and locally the user is a writer, so the count is never one. The rule is now that every implementer gets a worktree on either host. |
| `high` · the reviewer guidance is unactionable | Fixed. It now names `git show <commit>:<path>` or a detached worktree, and asks the reviewer to name the commit in its findings. |
| `medium` · the drift evidence is half false | Fixed. Checked review 5240258654 on #2: it reviewed `f56121b` and says nothing about the branch moving. The drift note appeared in its report to the caller, not in its findings. The text now says #1 recorded the commit it read and #2 did not notice its branch move, which is the case naming the commit protects against. |
| `medium` · wrong mechanism | Fixed. Two implementers work different sub-issues on different branches, so the one-branch-in-two-worktrees refusal never fires between them. It was never what made local parallel dispatch safe. The text now says what does happen: an index lock collision, or interleaved staging. |
| `medium` · harness worktrees conflict with the skill | Filed as #19 rather than fixed here. It predates this branch, the path claim needs confirming by dispatching an implementer rather than by reading configuration, and it changes what a run does rather than what a document says. |
| `low` · "Git raises nothing" is half true | Fixed by the rewrite above, which distinguishes the loud case from the silent one. |
| `low` · contradicted standing text | Resolved by the rewrite. The rule no longer carves out a host, so the sentences it disagreed with agree with it. |
| `low` · `docs/node_modules` is the real cloud gap, not sccache | Not addressed. It is about build isolation in a worktree, which this branch does not touch, and it deserves its own issue rather than a paragraph here. |

On the Complexity finding: taken. One rule replaced three cases, and it is the rule the repository already follows through `isolation: worktree`.

— claude-opus-5 · high

---

desertthunder:
## Adversarial review

Reviewed at `51c9faf`, read with `git show origin/agent/worktree-concurrency:<path>`; this container's checkout is on `claude/sleepy-carson-nhg3ex` at `f51217e`, so nothing here came from the working tree. Not my branch and not my earlier passes. One adversarial pass (5721204218) and one edit pass (5721236938) already ran, so this looks only at the text as it stands after `51c9faf`. Git claims were reproduced in scratch repositories on this box; the commit citations were checked against the object store and against the remote branch list.

### Findings

- `high` · `internal/thunderstorm.md:210` — the evidence, and the rule it argues for, do not survive a merge. `eb6e24f`, `6679d6d` (and `f56121b`, `7967ead` from the earlier discussion) are not objects in this repository: `git cat-file -t eb6e24f` → `fatal: Not a valid object name`. They are on no branch on the remote either — the branch list returns no `agent/*` branch for #1 or #2 — because `edge` takes pull requests as squashes (`ef58d61 … (#1)`, `3af907e … (#2)`) and `internal/thunderstorm.md:150` says "Delete the task branch when its pull request merges" → so the paragraph asks a later reader to trust two SHAs that reader cannot resolve, and the rule it justifies, "name the commit in the findings so a later reader knows what was reviewed", has the identical defect: the named SHA is unreachable the moment the pull request merges, which is the only moment anyone re-reads a review. A reviewer that follows this rule perfectly leaves a citation that is dead on arrival → pair the SHA with something that outlives the branch: the pull request number and a range against a surviving base (`origin/edge...<sha>`), or the review comment id, or say plainly that the SHA is good until merge and the squash commit on `edge` after. This is not the earlier `medium` on this paragraph, which was about whether the two reviews said what the text claims; the text was corrected and the citations are still unresolvable.

- `high` · `internal/thunderstorm.md:200-204` and `.claude/skills/thunderstorm/SKILL.md:43-46` — the shared-checkout failure mode is described as the wrong mechanism, and the worst case is missing. `SKILL.md:44` says two implementers in one checkout "hold different branches"; one checkout has one `HEAD`, so they cannot. What actually happens, reproduced:

  ```text
  A: git checkout -b agent/1 ; echo work > a1.txt ; git add a1.txt   # staged, not committed
  B: git checkout -b agent/2                                         # HEAD moves, A's index rides along
  B: git add b1.txt ; git commit -m "B: issue 2"
     → commit contents: a1.txt b1.txt
  A: git commit -m "A: issue 1"
     → landed on: agent/2
     → agent/1 tip still: aa81767 base
  ```

  So the silent case is not "interleaved staging": it is total branch capture. A's work is inside B's pull request, `agent/1` never leaves `origin/edge`, and A's own pull request is empty. `push-verified.sh` does not catch it — both pushes have a symbolic ref and both land — and the only place it surfaces is that script's last line naming `agent/2` where the run expected `agent/1`. This is also the first thing a shared checkout would hit, because creating the branch is a dispatched implementer's first act (`thunderstorm/SKILL.md:51`) → say `HEAD` capture rather than staging interleave, and drop "they hold different branches"; the index lock is the rarer case, not the representative one.

- `medium` · `.claude/skills/worktree/SKILL.md:34-36` — the reviewer rule is written in the one document a reviewer never opens. `.claude/agents/reviewer.md` and `.claude/agents/adversarial-reviewer.md` both say only "Use the `review` skill", and `review/SKILL.md` still prescribes `gh pr diff <n>`, `git log --oneline origin/edge..<branch>`, MCP `get_diff`, and "Read the changed files around the diff" — every one of which follows the moving ref or the working tree. Neither agent file nor the review skill mentions `git show`, a commit, or the `worktree` skill, and that skill's own description triggers on "starting work on an issue, running parallel agents, or cleaning up after a run", so a reviewer has no reason to load it; `internal/thunderstorm.md`'s "Who gets one" sits under `## Worktrees`, which it also never opens. `review/SKILL.md:77` fixes the finding format and `:97` the signature, and neither has a slot for the commit, so "name the commit in the findings" asks for a field the format does not define. Failing case, live in this pass: the container checkout is `claude/sleepy-carson-nhg3ex` at `f51217e`; a reviewer using `Read` would have reviewed `edge` and reported on a diff that is not this pull request's → put the read discipline and the commit line in `review/SKILL.md`, add the commit to its signature block, and leave `worktree/SKILL.md` pointing at it. Flagging the stop condition: this is the cause behind the earlier pass's second `high` surviving one edit pass in a new form — the wording was made actionable, the placement was not. If it survives the next edit pass unchanged, escalate rather than posting it a third time.

- `medium` · `.claude/skills/worktree/SKILL.md:34` and `internal/thunderstorm.md:208` — `git show <commit>:<path>` is given with no fetch, and in a cloud container the object is not there. This container's clone is shallow (`git rev-parse --is-shallow-repository` → `true`, `edge` truncated at 50 commits, grafted at `7cdec7f` and `9ba6e1b`) and at session start carried only `origin/edge`, `origin/main` and the session's own branch; this pass had to run `git fetch origin agent/worktree-concurrency`, which printed `* [new branch]`. A reviewer following the line verbatim gets `fatal: invalid object name` — reproduced exactly, with `git show eb6e24f:internal/thunderstorm.md`. The Create block two sections up opens with `git fetch origin` for precisely this reason; the reviewer paragraph has no equivalent → open it with `git fetch origin <branch>`. Note also that `review/SKILL.md`'s mandated `git log --oneline origin/edge..<branch>` has no merge base when a branch forked earlier than the graft point; not observed on this branch, but the truncation is real and the cloud is now the host this document says behaves "the same as local".

- `low` · `.claude/skills/worktree/SKILL.md:24-26` — "A cloud session renames its branch to match this convention before the first push" is left over from the sentence this diff deleted, and now contradicts the paragraph two lines below it. It only ever made sense as the exception clause of "A cloud session skips the worktree". Under the new rule the cloud worktree is born on `agent/<issue>` by `git worktree add … -b agent/<issue>` at `:18`, so there is nothing to rename; read on its own the sentence says a cloud session works in its container checkout and renames the harness branch, which `:28-32` forbids → delete it, or scope it explicitly to a session with no dispatched implementer, where the container checkout is the only tree.

- `low` · `internal/thunderstorm.md:206` — "It does not write" is the premise for giving a reviewer no isolation, and it is false: `review/SKILL.md:101` posts with `gh pr comment <n> --body-file <file>`, and the reviewer's tool set includes `Bash`. `.gitignore` is not allowlist-shaped (it names `/target`, `.thndrs`, `.env`, `.zed` and negations), so a findings file written into a checkout the reviewer now shares with a writer is untracked and visible, and `git status` in that tree attributes it to nobody → say the reviewer writes nothing *into the tree* and name the scratch location, rather than claiming it does not write.

- `nit` · `internal/thunderstorm.md:32-35` — both rows of the Worktrees column now say "see [Worktrees](#worktrees)", so the column carries no information and the table's reason for existing is the transport column alone → drop the column and let the sentence above the table carry "the protocol is the same".

### Complexity

- One rule is now stated three times — `thunderstorm/SKILL.md:43-46`, `worktree/SKILL.md:28-32`, `internal/thunderstorm.md:194-204` — and the copies have already drifted: the skill copy says the two implementers "hold different branches", the internal copy says they "work different sub-issues on different branches", and the first is impossible in the situation both describe. `internal/thunderstorm.md:166`, rule 2 under "Recording a failure mode", asks for one or two sentences in the skill that owns the work, not a paragraph in each document that touches it → keep the mechanism in `worktree/SKILL.md`, and let the other two cite it.
- Rule 1 of that same list (`:164`) — "Make it impossible, or make it fail loudly. A check that runs beats a rule that has to be remembered" — is available here, and the pull request's own "Not covered" concedes the rule enforces nothing. The capture in the second finding is caught by one argument: have the implementer pass its claimed branch to `push-verified.sh` and have the script refuse when `git symbolic-ref --short HEAD` does not match. That is three lines in a script that already exists for the same class of failure, it fires at the point the damage becomes visible, and by rule 3 it lets most of this prose go. A `git rev-parse --show-toplevel` assertion against the main checkout covers the rest.

### Checked and found clean

- Worktree creation inside a shallow clone works and the worktree is shallow with it (`--depth 2` clone, `git worktree add … -b agent/99 origin/edge` → branch created, full tree present), so requiring worktrees in cloud containers costs nothing on that axis. I expected this to break and it does not.
- `git worktree add --detach <dir> <sha>` succeeds beside an existing branch worktree, so the reviewer's second option is real.
- `push-verified.sh` refuses a detached `HEAD` outright, so a reviewer taking the detached worktree cannot push from it by accident.
- The diff touches no `.rs`, no manifest, and no workflow, as the body says.

— claude-opus-5 · high

---

desertthunder:
Findings addressed in `95a7d89`. Disclosure: I wrote the adversarial pass I am answering here, in this session, so a reader should weigh these outcomes accordingly — a pass that grades its own homework is not a review.

| Finding | Outcome |
| --- | --- |
| `high` · the citations and the rule do not survive a merge | Fixed. The two SHAs are gone rather than re-cited: they were reachable from nothing, which was the finding. The rule now asks for the commit **and** the pull request number, and says why — the branch is deleted at merge and the commit goes with it. The anecdote went too. I drafted a replacement reading "both adversarial passes read a branch that moved under them", caught that the half about #2 was never established (the commit that could have moved under it is not the one the earlier discussion named), and dropped it rather than trade one unverifiable citation for another. The mechanism does not need the story. |
| `high` · wrong mechanism, worst case missing | Fixed in all three copies. Two implementers in one checkout share one `HEAD` and *cannot* hold a branch each; the second to create its branch takes the first's staged work onto it, the first's later commits land there too, and the first's branch never leaves `origin/edge`. The text now says that, says the one-branch refusal cannot fire because there is only one worktree, and says `push-verified.sh` passes both pushes. The index lock is demoted to the rarer, louder case. |
| `medium` · the rule is written where no reviewer reads it | Fixed, as the placement problem it is. The read discipline is now a section of `review/SKILL.md` — the document both reviewer agents load — covering the fetch, `git show <commit>:<path>`, the three-dot diff, the detached worktree, and an explicit "do not read the change with `Read`, `Grep`, or `Glob`, they read the working tree". `worktree/SKILL.md` and `internal/thunderstorm.md` now point at it instead of restating it. The commit line went into the Post section, which is where the format that lacked a slot for it lives. |
| `medium` · no fetch step, and the cloud object is absent | Fixed. The block opens with `git fetch origin <branch>` and says the fetch is not optional, naming why: a cloud container clones a few branches at a shallow depth, so the branch under review is usually absent and `git show` fails on a commit the repository has never had. |
| `low` · vestigial cloud-rename sentence | Fixed. It now scopes itself to a session that took a harness-generated name and has no worktree yet, and says plainly that a worktree made by this skill is born with the right name and needs no rename. |
| `low` · "It does not write" is false | Fixed. The claim is now "writes nothing into the tree", and `review/SKILL.md` says where the comment body goes: the session's scratch directory, never inside the repository, because a findings file left in a shared checkout is untracked work belonging to nobody. |
| `nit` · vacuous Worktrees column | Fixed. The column is gone and one sentence under the table carries it. |
| Complexity · one rule stated three times, already drifting | Fixed. The mechanism lives in `worktree/SKILL.md`; `thunderstorm/SKILL.md` keeps three lines and a pointer, `internal/thunderstorm.md` keeps the spec-level statement and defers the reading discipline to the `review` skill. |
| Complexity · make it loud instead of writing it down | Deferred to #21. The guard is right — `push-verified.sh` taking the branch the run claimed and refusing on mismatch — but it is a script change with a test beside `sync-labels-test.py` and a CI job, not a documentation change. #21 carries the reproduction, the acceptance criteria, and the note that it makes the loss loud without preventing it. The prose it would replace is named there too, so the trim happens on that branch rather than being forgotten. |

## Verification

- `check-commit-message.py --range HEAD~1..HEAD` — clean.
- `check-frontmatter.py internal/` — `thunderstorm.md` passes. It reports 8 other files without frontmatter; those are the files this branch is behind `edge` on, fixed by #10, and untouched here.
- No `.rs`, no manifest, nothing under `docs/`, so the Rust gates and the docs build have nothing to run.
- Edits were made in a worktree at `../thndrs-worktrees/11`, not in the container checkout, which sits on another branch.

— claude-opus-5 · high

---

Issue #12, opened 2026-09-17
Make the thunderstorm workflow ready for feature work


desertthunder:
## Goal

The loop can be started by a human, worked by an agent from either host, and trusted without anyone remembering an unwritten rule. Everything here is workflow readiness, not product work: it is what stands between the current state and running `/impl` on a feature issue without hesitation.

## Done when

Every sub-issue is terminal, and a `/thunderstorm` run over a feature epic completes without a participant needing to consult this epic to know what to do.

## Planned sub-issues

- [ ] #3 — tag a release, so the release skill has a previous version to read
- [ ] #4 — assign OpenCode Go and Cursor roles, or drop the columns
- [ ] #5 — give cloud sessions their own GitHub identity
- [ ] #6 — close the dispatch gap on the labels apply job
- [ ] #7 — decide whether CI pins the Rust toolchain
- [x] #8 — add frontmatter to the internal files that predate it

#5 and #7 are decisions rather than implementations: their done-condition is a recorded decision with its reason, not a code change.

## Source

`internal/thunderstorm.md` (`01M2PWX233GKXE5M9SPTNTGN0D`), and the two pull requests that built the cloud path: #1 and #2. Each sub-issue cites the same identifier.

## Not in this epic

Product work. `internal/BUGS.md` holds seven interface items and `internal/docs-update.md` has eleven documentation tasks outstanding; both are their own epics, and most of the BUGS list needs a design pass before it has criteria anyone could verify.

Enforcement of the workflow's own rules is also excluded. It is a real gap — the board invariants are stated as musts and checked by nothing — but it is a design question first, not a task.

PR #13, opened 2026-09-17
docs: record where enforcement helps and where it does not

desertthunder:
## What

One idea file, `internal/ideas/enforcing-workflow-rules.md` (`01M2RH6SW2KZHH0V3BRW0NVMKT`), from a `/r-d` session on whether the workflow needs to enforce its own rules.

## Why

Thirteen sentences across `internal/thunderstorm.md` and `.claude/skills/github-board/SKILL.md` are phrased as a prohibition or an obligation. Two workflows exist and neither reads an issue, so the board — which `github-board` says misroutes the loop when an issue carries two status labels — is checked by nothing.

The decisions recorded:

- Enforcement is three levels, not one. Prevention, detection, blocking. Only blocking creates bypass pressure, and treating them as one word is how a rule that deserved a whisper ends up stopping work.
- Enforce the board invariants at **detection**: one `status:*` per issue, an epic carries none, `status:claimed` has an assignee, `status:blocked` has a reason. Machine-checkable without judgment, silent when violated, and the board routes every run.
- A board check does not block a merge. The board and the code are separate systems.
- Do **not** enforce the process rules — worktree per writer, the review sequence, model assignments. None is observable from outside the session that performed it, so a check would inspect artifacts a careless run produces just as well as a careful one.
- A guard that misfires is worse than the rule written down, with the evidence in this repository: the commit-msg hook shipped rejecting `git commit -v` and refusing `git merge`, and the only way past either was `--no-verify`, which also skips the check for the next message.
- Enforcement and review are not substitutes. The two costliest errors while building the cloud path were claims made without the verification they implied — a push reported from an exit code rather than the remote, and a repository-wide claim made from a search of three directories. No check catches those. An adversarial reviewer did.

## Verification

Claims in the file were checked against the repository rather than recalled: the workflow list and their triggers, that no trigger fires on an issue event, the count of must-phrased sentences, and the commit-msg defects as reported in the adversarial review on #2.

Prose only, so no gate is exercised beyond the commit-message job.

## Not covered

- No check is implemented. The file is an idea, and the rubber-duck skill says an idea becomes work when it has criteria someone could verify. The board check has a shape but not a done-condition.
- The open question is whether violations actually occur. Six issues and one epic exist, all conforming, none worked by a run yet. If twenty pass through clean, the check is ceremony and the file says to delete it rather than keep it.
- `push-verified.sh` stays self-invoked. It prevents exactly the failure you have when you are not thinking about it, and a pre-push hook cannot cover it, because git does not run one when the push has no ref to update — which is the case it exists for.

---


internal/ideas/enforcing-workflow-rules.md
```diff
@@ -0,0 +1,94 @@
...
+# Enforcing the workflow's own rules
+
+## Problem
+
+The workflow states rules as musts and checks almost none of them.
+`internal/thunderstorm.md` and `.claude/skills/github-board/SKILL.md` carry
+thirteen sentences between them phrased as a prohibition or an obligation. Two
+workflows exist, `ci.yml` and `labels.yml`, and neither reads an issue.
+
+The board is the part that matters. Status is the state, and `github-board`
```
desertthunder: We need some guardrail in the CLAUDE/AGENTS.md file that ensures that no AI/Claude-isms or tropes get through.

desertthunder:
Standard review pass.

```text
medium · internal/ideas/enforcing-workflow-rules.md:11 — "neither reads an issue" is
false: labels.yml holds issues: read on the dry-run job and issues: write on the apply
job, and sync-labels.py:183 queries repos/{owner}/{repo}/issues on every run to count
what carries a label → the sentence sits in Problem where it carries the argument, and
the file's own Sources section states the accurate version ("the absence of any trigger
on an issue event"), so the document contradicts itself → say neither workflow is
triggered by an issue event and neither checks issue state.

medium · internal/ideas/enforcing-workflow-rules.md:10 — "thirteen sentences ... phrased
as a prohibition or an obligation" does not survive counting: sentences matching that
description across both files number 9 → the 13 came from grep -c over the two files,
which counts lines containing never/must/do-not, inflated by multi-line sentences and by
"must" in non-obligation use; a precise number from a method that does not measure it,
in a document whose argument is that unverified claims are what enforcement cannot catch
→ state 9, or drop the count and name the two files.

low · internal/ideas/enforcing-workflow-rules.md:29 — the table column reads "Caught by",
and the row "Two false claims that a push landed | A check that had to be invoked"
credits push-verified.sh, which did not exist when those claims were made and was written
afterwards as the fix → nothing caught them at the time; the absence of a CI run for the
commits did, later → say so, which is a sharper point for the file's own argument than
crediting a check that was not there.

nit · internal/ideas/enforcing-workflow-rules.md:69 — "Six issues and one epic exist as
of this writing" is eight issues now, #9 having been filed since → the hedge holds, but a
count ages and the claim it supports does not need one → drop the number.
```

Checked and clean: the file matches the `rubber-duck` template exactly, in order, with a well-formed ULID. The pre-push claim is accurate and was reproduced. The commit-msg defects are cited correctly from the adversarial pass on #2. Prose wraps clean.

The conformance claim in `Open` is stronger now than when it was written: all eight open issues carry exactly one `status:*` with one each of `type:`, `area:` and `risk:`, and #12 carries `kind:epic` with no status — including the issues this session did not file.

— claude-opus-5 · high

---

desertthunder:
Adversarial review pass. The head already carries `1e2ea11`, so this reads the corrected text; two of those corrections do not hold (findings 4 and 8).

```text
high · internal/ideas/enforcing-workflow-rules.md:49 — "`status:claimed` has an assignee"
is listed as a board invariant, but `github-board/SKILL.md:107` creates that exact state
deliberately: "A session killed between the two calls leaves `status:claimed` with no
assignee, which the 24-hour rule returns to the queue", and again at :125 for the claim
race → a detection check built from this list fires on every claim in flight (the window
between the two calls, visible in #8's timeline as two events at 20:05:57Z) and on every
interrupted claim for up to 24 hours, states the skill designed for and tolerates; that is
the file's own worst outcome at line 62, a guard that misfires → state it as the skill
does, with the grace period: no assignee for more than 24 hours, and no PR.

high · internal/ideas/enforcing-workflow-rules.md:56 — "Do not enforce the process rules:
a worktree per concurrent writer, the review sequence, the model assignments. None is
observable from outside the session that performed it" is false for two of the three.
`internal/thunderstorm.md` requires "Every comment ends with a signature naming the model
and its reasoning level", `internal/models.md:8` repeats it, and the implementer's model
is on the PR in the `Co-Authored-By:` trailer both commits on this branch carry → the
review sequence and "the implementer and the reviewer never share a model in one run" are
both readable from a PR's comment and commit record without entering the session, so the
decision rules out the check the signature convention exists to make possible → drop those
two from the list, or say why the record is insufficient rather than absent.

high · internal/ideas/enforcing-workflow-rules.md:84 — "a pre-push hook is not run when the
push has no ref to update" is false. Reproduced on git 2.43.0: a no-op push prints
`Everything up-to-date` and the hook still runs, with empty stdin; a hook that refuses on
empty stdin blocks it, `error: failed to push some refs`, exit 1 → the only Open item about
automating `push-verified.sh` is closed on a mechanism claim that does not hold, so the
obvious fix is ruled out before anyone tries it. Inherited, not introduced: the same claim
sits in `internal/thunderstorm.md:155` and in `push-verified.sh`'s header → verify against
a running git and rewrite; if the conclusion changes this becomes a candidate check.

high · internal/ideas/enforcing-workflow-rules.md:74 — "Every open issue conforms as of this
writing, and none has been worked by a run" was false when written. #8's timeline shows
`status:queued` 20:01:46Z, `status:claimed` plus `assigned` 20:05:57Z, `status:review`
20:13:53Z, PR #10 opened 20:13:49Z — a full claim-to-review pass 21 minutes before the
first commit on this branch and 33 before the commit that edited this very sentence for
staleness → the premise of the whole Open item is that the board has no history, and the
counter-example was on the board and visible while the sentence was being corrected →
replace with what the record shows: one issue has been through a run, and what it violated.

medium · internal/ideas/enforcing-workflow-rules.md:48 — the four-invariant list is a hand
copy of rules defined elsewhere and has already diverged at birth. `github-board` says an
epic "has no owner and no `status:*` label" (the list keeps the label half, drops the
owner); "an issue that already carries an assignee is not claimable" (omitted); "assignees
is exactly your login and nothing else" (omitted); and Rules, "Do not close an issue to
express any state other than `dropped`" (omitted). That last is not hypothetical: #8 is
closed right now carrying `status:review`, auto-closed 21:09:50Z one second after #10
merged, because #10's body opens with `Closes #8`. It never reached `status:verify`, and
the check as listed would pass it → the first board violation this repository has produced
is of a rule the list omits, by a mechanism nobody in the loop performs deliberately, and
`thunderstorm.md` warns about exactly this copy-drift → name the source sections as the
definition, and scope the check to closed issues as well as open.

medium · internal/ideas/enforcing-workflow-rules.md:22 — "the rules that had a check were
the ones that held" and the first table row credit the commit-msg hook, but the check has
a hole where messages actually land. The repo's own checker over `edge`
(`check-commit-message.py --range <root>..origin/edge --warn`) fails two of the three
squash commits merged since the hook shipped: `3af907e` subject 62 chars, `f51217e` 65,
over the 59 limit. GitHub composes the squash message server-side and appends ` (#NN)`, so
`commit-msg` never sees it and the CI `commits` job cannot either — it runs over the
branch's own commits and reports without failing → the example Decisions leans on has a
permanent record of not working, on the merge path used by every PR → qualify the row, or
make it the second Open item; a squash-subject check is prevention-level and cheap.

medium · internal/ideas/enforcing-workflow-rules.md:94 — Sources credits "The adversarial
review on #2" with the commit-msg hook defects. Both are on #1, at 2026-09-17T16:06:14Z
(`.githooks/commit-msg:12`, the scissors-line grading and the `git merge` rejection), under
a review body reading "Adversarial pass on `agent/git-hygiene`"; #2's adversarial review is
entirely about `sync-labels` and mentions the hook nowhere. The hook shipped in `ef58d61`
(#1) → Sources exists to make the trail readable and points at the wrong PR. This also
means the standard pass's check that the defects are "cited correctly from the adversarial
pass on #2" is itself wrong → change `#2` to `#1`.

medium · internal/ideas/enforcing-workflow-rules.md:13 — "nine sentences between them
phrased as a prohibition or an obligation" does not survive recounting either, so the
correction swapped one unreproducible number for another. Neither document contains the
word "must" at all (`grep -ci '\bmust\b'` returns 0 for both). Splitting both files into
sentences with fences and table rows removed and counting `never|do not|cannot|must|only`
gives 6 and 12 — 18, a floor, excluding bare imperatives that are plainly obligations
("Remove the worktree when the run ends", "One writer at a time", "Dispatch on `edge` and
nothing else") → the file's thesis turned on itself, a confident count no reader can
reproduce, in the sentence that opens the argument → drop the number, or state the counting
rule inline so it can be re-run.

low · internal/ideas/enforcing-workflow-rules.md:75 — the Open section's exit condition has
no recordable done-condition. "If twenty pass through without a violation" lost its noun in
the correction commit (the antecedent was "twenty issues"; the nearest plural is now "board
violations"), and nothing counts issues that have passed through a run, so no one can ever
say the twenty happened. A detection check that posts a comment is not a durable record of
a clean pass either. The second Open item names a consideration but no settling condition,
where the `rubber-duck` template asks for "what would settle it" → restore the noun and
name where the count lives, or replace it with a dated revisit.

low · internal/ideas/enforcing-workflow-rules.md:48 — the decision to add a board check
collides with `internal/thunderstorm.md:168`, "Delete the prose once a check covers it.
Guidance that describes a failure something else now catches is read as optional and trains
people to skim." Taken literally, shipping the detection check obliges deleting "Never add
a status label without removing the previous one", "A block with no reason label is not a
block", and the epic sentence → the rule then lives only in a script whose output is an
advisory comment, which the file itself calls weak ("Detection assumes someone reads the
comment"); step 3 was written for blocking checks and has no carve-out for detection → say
step 3 applies at prevention and blocking, not detection, or record why this is an exception.
```

Checked and clean: frontmatter `name` matches the filename, `last_updated` matches both commit dates, and the ULID is 26 valid Crockford characters colliding with nothing in the tree. Section order matches the `rubber-duck` template exactly, and the file stays inside that skill's "Do not" list. `.github/workflows/` holds exactly `ci.yml` and `labels.yml` and neither triggers on an issue event, so the corrected lines 14-16 are accurate. "pick the wrong transition" quotes `github-board` verbatim and in context. Two of the four invariants do match their sources exactly: one `status:*` per issue, and `status:blocked` carrying a `blocked:*` reason. Both open epics (#12, #16) carry `kind:epic` and no status. "A board check does not block a merge" contradicts nothing — no document gates a merge on board state. `push-verified.sh` does what the file says. Both commit-msg defects cited are genuinely fixed on this branch, so the past-tense framing at line 62 is right even though the citation points at the wrong PR.

The four `high` findings were re-verified independently of the reviewing agent: the no-op pre-push hook was reproduced in a scratch repo on git 2.43.0, the `#1` citation against that PR's review comments, the claimed-without-assignee tolerance against `github-board/SKILL.md:107` and `:125`, and #8's closed-carrying-`status:review` state against the issue itself.

— claude-opus-5 · high

---

desertthunder:
Edit pass on the adversarial findings. Pushed as `09d12a1`, confirmed on the remote (`origin/agent/enforcement-idea` at `09d12a15`).

| # | Finding | Outcome |
| - | ------- | ------- |
| 1 | `status:claimed` has an assignee | Fixed |
| 2 | Process rules "not observable" | Fixed, on different grounds |
| 3 | Pre-push hook claim | Fixed; two inherited copies deferred |
| 4 | "None has been worked by a run" | Fixed |
| 5 | Invariant list copy-drift | Fixed in part; close-rule half **withdrawn** |
| 6 | Commit-msg hook's squash hole | Fixed |
| 7 | `#2` should be `#1` | Fixed |
| 8 | "Nine sentences" | Fixed |
| 9 | Open exit condition has no noun | Fixed |
| 10 | Collision with step 3 | Fixed |

**Withdrawn — finding 5's second half.** I reported #8's auto-close as "the first board violation this repository has produced". That was wrong, and the repository owner has said so directly: `Closes #N` closing its issue on merge is what the keyword is for, and auto-closing is fine. The rule it cited, `github-board/SKILL.md:184` ("Do not close an issue to express any state other than `dropped`"), is the defect. Honouring it would mean banning closing keywords or reopening issues by hand after every merge. The recommendation attached to that finding — scope the check to closed issues — would have built a guard firing on every issue the workflow completes, which is the exact failure the file argues against. The file now records the close rule as one to delete rather than check. **`github-board/SKILL.md:184` is still there and still wrong; fixing it is a separate change and not in this PR.**

The copy-drift half stands and is fixed: the list now points at the Statuses table in `thunderstorm.md` and the Claim and Rules sections of `github-board` instead of restating four invariants from memory.

**Finding 2, fixed on different grounds.** The premise was false — the signature convention and the `Co-Authored-By:` trailer do put the review sequence and model assignments on the PR. But the decision not to enforce them survives for a better reason: those artifacts are self-reported by the session under test, so a check over them reads a claim rather than a fact, and a run that skipped a pass is the run least likely to say so. That is the same class as the two false push claims the file already uses, so it now argues from that instead of from a false claim about observability. The worktree rule genuinely is invisible from outside and stays as it was.

**Finding 3, and what is deferred.** Reproduced against a scratch remote on git 2.43.0, so the file now states verified behaviour rather than the reverse:

- A no-op push **does** run `pre-push`, with empty stdin, and prints `Everything up-to-date`.
- A hook refusing on empty stdin fails the push: `error: failed to push some refs`, exit 1.
- Detached `HEAD` under the default `push.default=simple` does not reach that path at all — git refuses with `fatal: You are not currently on a branch`. `Everything up-to-date` from a detached `HEAD` needed `push.default=matching`.

So a pre-push hook is a candidate for carrying `push-verified.sh`, not an impossibility, and that Open item now says so. **Deferred:** the same wrong reason is recorded in `internal/thunderstorm.md:155` and in the header comment of `.claude/scripts/push-verified.sh`. Both are outside this PR's one file. The file names them; they need their own change.

**Finding 6.** Verified independently: of the three squashes merged since the hook shipped in `ef58d61`, `3af907e` has a 62-character subject and `f51217e` a 65, against the 59 the hook enforces. GitHub composes the squash message server-side, so neither the hook nor the CI `commits` job ever sees it. The table row is qualified to "locally" and the gap is stated in prose.

**Finding 8.** Both counts were unreproducible because the premise was wrong: neither source file contains the word "must" at all, and the rules are carried in the imperative. The number is gone rather than replaced with a third one.

One correction to my own adversarial comment: I wrote that `thunderstorm.md` warns about copy-drift, citing a quote my first grep did not find. The quote is real — "duplicating its sub-issues' state on the parent gives that copy somewhere to drift" — it spans lines 78-79, so a single-line search missed it. The claim held; my verification of it initially did not.

Verification: `prettier --check` clean (no change needed), `check-commit-message.py --range origin/edge..HEAD` reports 3 commits clean, push confirmed by comparing the remote ref rather than by the exit code. No Rust or `docs/` files changed, so per `CLAUDE.md` the cargo gates and the docs build do not apply.

Not reviewing this revision — that is a separate pass.

— claude-opus-5 · high

---

desertthunder:
Edit pass. One finding, raised by the repository owner and not by a review pass.

| Finding | Outcome |
| ------- | ------- |
| The previous revision rewrote prose without the `writing-docs` skill | Fixed |

`CLAUDE.md` routes prose through `writing-docs`, and `commits-and-prs` says so again for commit messages. The previous revision rewrote roughly a hundred lines of this file and loaded neither. Pushed as `8a4e928`, confirmed on the remote (`origin/agent/enforcement-idea` at `8a4e928c`).

Measured against the skill after loading it, the prose I had added carried:

| Tell | Before | After |
| ---- | ------ | ----- |
| Sentences over 25 words | 14 | 6 |
| Em dashes | 4, three in one section | 0 |
| `X, not Y` reframes | 4 | 0 |
| Phrases from the tells table | 1 (`the point`) | 0 |
| Lines over 80 columns | 1 | 0 |
| Paragraphs over 5 sentences | 2 | 0 |

The six sentences still over 25 words are two of the original author's, two source citations, and two carrying three-part enumerations that do not split without repeating the subject. The skill treats the limits as warnings to restructure against, not as hard caps, and these read cleanly.

No fact, number, qualifier, or identifier changed. I checked each one is still present rather than assuming the rewrite preserved them: the commit SHAs and their subject lengths, the 59-character limit, the 24 hours, the git version and both `push.default` values, and every issue and pull request reference. `prettier --check` passes.

Worth recording against this file's own argument, since it is a data point and not a favourable one. Loading the skill is a process rule, and the Decisions section says not to enforce process rules because a check over their artifacts reads what the session claims rather than what it did. That holds here: nothing in the diff, the commit, or CI showed the skill had been skipped. What caught it was a person reading the prose. That is the third entry for the "caught by a reviewer" row, and the second where the failure was a rule this repository had already written down.

— claude-opus-5 · high

---

internal/ideas/enforcing-workflow-rules.md
```diff
@@ -0,0 +1,170 @@
...
+issue event and neither checks issue state, though `labels.yml` reads issues to
+count what carries a label.
+
+Every run reads the board to pick its next issue and writes the transition
+back. Status is the state, and `github-board` says two status labels on one
+issue make the loop "pick the wrong transition". That is silent misrouting, not
+a visible failure, and nothing would notice it.
+
+The gap is not theoretical. Over 2026-09-17, building the cloud path, the rules
+that had a check were the ones that held:
```
desertthunder, line 30: This is slop

internal/ideas/enforcing-workflow-rules.md
```diff
@@ -0,0 +1,170 @@
...
+| What went wrong                      | Caught by                        |
+| ------------------------------------ | -------------------------------- |
+| Five commit messages over the limits | The commit-msg hook, locally     |
+| Doubled blank lines from an edit     | `cargo fmt` in CI                |
+| A parse failure reported as success  | The label sync test, once it ran |
+| Two false claims that a push landed  | Nothing, until CI showed no run  |
+| A false claim that a rename was done | An adversarial reviewer          |
+
+The first row overstates its check. The hook grades what a committer writes
+locally. GitHub composes a squash message server-side and appends ` (#NN)`, so
```
desertthunder, line 41: also slop

PR #14, opened 2026-09-17
docs: record screenshot-driven TUI verification idea

desertthunder:
## What

Adds `internal/ideas/tui-screenshot-verification.md`, the output of an `/r-d`
session on capturing TUI frames at pull request time with tmux and freeze. It
records decisions. No code, skill, or workflow changes.

## Why

No review pass runs the binary. All three read the diff, and running behavior
is confirmed at `status:verify`, after the branch reaches `edge`, by one
person. Starting the application once during the session printed `Skill skipped
(session-start-hook): name "startup-hook-skill" must match parent directory`,
which no reading of the diff that introduced it would have found.

Three obstacles stand in the way of capturing frames today.
`docs/.../tui-qa.md` calls `capture-pane` with no `-S`, so it reads the visible
pane while `runtime/terminal.rs:170` moves finished rows into terminal history
through `insert_before`. The same document waits with `sleep 0.2`. No session
fixtures exist, so any capture past the composer costs a live model call.

## Verification

- `python3 .claude/scripts/check-commit-message.py --range HEAD~2..HEAD` —
  2 messages checked, all clean.
- `awk 'length > 80'` over the file — no output.
- No Rust changed, so the `cargo` gate did not run. `CLAUDE.md` exempts
  `internal/` from `pnpm --dir docs build`, which also did not run.

The claims in the file were measured during the session. `go install
github.com/charmbracelet/freeze@latest` resolved v0.2.2 and rendered a live
`tmux capture-pane` of thndrs to PNG in 1.5s. Freeze dropped `\e[3m`, `\e[2m`,
`\e[7m`, and `\e[41m` while rendering bold, underline, 256-color and truecolor.
`npm view` resolved `opencode-ai`, `@openai/codex`, and `@sourcegraph/amp`,
but not `@factory-ai/cli`.

## Not covered

The file records an idea with no acceptance criteria yet, so no issue is
linked. It leaves four questions open: which `pi` and `grok` packages are
meant, whether comparative captures run in the container or locally, how
fixture session JSONL gets authored without leaking credentials or paths, and
whether motion needs VHS. The freeze measurements come from v0.2.2 in the
Claude Code cloud container and were not repeated on macOS. Streaming, spinner
behavior, and cursor movement stay unverified.

https://claude.ai/code/session_013eypnRUw26FMvUfqr5eL1v


internal/ideas/tui-screenshot-verification.md
```diff
@@ -0,0 +1,139 @@
...
+  history, and the cleanup block that confirms with `has-session`. Their
+  repository layout and `Herdr` workflow no longer apply.
+- Keep the polish rubric from `tui-design/references/harness-patterns.md` as
+  it stands: hierarchy, composition, rhythm, restraint, legibility, state
+  craft, stability, and voice, each scored 0 to 2, any zero blocking
+  completion.
+
+## Open
+
+Which `pi` and `grok` packages are meant. The registry holds `@badlogic/pi`
```
desertthunder: Grok: https://github.com/xai-org/grok-build
Pi: https://github.com/earendil-works/pi

desertthunder: Fixed in `54b74ac`.

None of the four packages I guessed were the right projects, so the open
question is answered and removed rather than narrowed. `@earendil-works/pi-coding-agent`
resolves at 0.85.1 and joins the three that already installed from npm here,
which makes four of the six runnable in the same tmux geometry.
`xai-org/grok-build` stays source-only alongside `@factory-ai/cli`, which does
not resolve.

One thing the link turned up that changes a decision rather than a fact: Pi
publishes its terminal UI separately as `@earendil-works/pi-tui`, a library with
differential rendering. Its layout decisions can be read in source instead of
inferred from a captured frame, so I added a line saying to read it next to its
harness. That is the better evidence for the patterns the rubric measures.

Still open, unchanged by this: container versus local capture, how fixture
session JSONL gets authored, and whether motion needs VHS.

— claude-opus-5

---

PR #15, opened 2026-09-17
feat: add the spec and decomposition stages

desertthunder:
## What

Two skills and their commands, plus the stage list that connects them.

- `/spec-ify`, `/specify` → `specify`. One design to `internal/features/<name>/plan.md`, written only when a decision is missing.
- `/decomp`, `/decompose` → `decompose`. Cuts an idea, spec, or finding into one epic and its sub-issues.

`internal/thunderstorm.md` gains a **Stages** section naming the whole path, and `internal/features/` finally appears in the layout table.

## Why

**The spec stage already existed and nothing owned it.** Four skills referred to it — `rubber-duck` describes itself as running "before any code or spec exists" and refuses to expand an idea into one, `implement` says "read the spec the issue points at", `github-board` defines `blocked:spec` — while `internal/features/` has carried five feature designs the whole time and did not appear in the layout table.

**Nothing owned decomposition either.** The rules for turning a document into an epic and sub-issues were spread across four files: cite the source identifier, one epic with its own stop rule, criteria checkable without judgment, a scope bound, no status on an epic, the label triple. The last decomposition was done by hand against those scattered rules and still had to be told about citing the source document.

`github-board` is not this stage. It knows how to claim an issue and move a status; it decides nothing about which issues should exist.

## The decisions worth arguing with

**A spec is conditional, not a pipeline step.** The test is one question: can you write a sub-issue's stop rule without deciding anything else first? If yes, a spec adds nothing. The six issues filed today came straight from findings and have checkable criteria; a spec for each would be ceremony. "Give tool activities distinct semantic kinds" cannot produce criteria until someone settles the kinds, and five issues each inventing their own answer is the failure a spec prevents.

**A new track gets no `tasks.md`.** The existing ones are a checklist in a document — a second copy of the board's state that nothing updates and that is wrong the first time an issue closes. That is the same failure removed from `internal/thunderstorm.md` earlier today, when its `Not wired up yet` list had two of five bullets already false. The rule lives in `specify`, where the writer who would add one reads it; the features README does not explain the legacy tracks, because they are going away.

**A spec is reviewed like anything else.** It sets the vocabulary for every issue cut from it, so it goes through a pull request and `/rev`. That is the reason `rubber-duck` refuses to expand an idea into a spec in the first place.

**A filed epic is only reachable through the sub-issue relation.** Added after the adversarial pass. A run finds its work by asking GitHub for an epic's children, so the list in the epic's body is a plan and never the link. The epic template stopped asking for a checklist beside the relation for the same reason `tasks.md` went.

**`What is not done yet` is no longer a section.** Its two rules moved into **File conventions**: an issue cites the identifier of the document it came from, and no document carries a backlog because the board does.

## Verification

Claims checked against the repository rather than recalled: the four skills referencing a spec stage, the five feature directories with `plan.md` and `tasks.md`, the absence of an `internal/features/` row in the layout table, and `mcp/tasks.md` carrying checked boxes.

Prose, skill files, and `.github/ISSUE_TEMPLATE/epic.yml`, whose stop-rule example and sub-issue field contradicted the skill that cites them. No script, workflow, or manifest changes, and no Rust or `docs/`, so the cargo and `pnpm --dir docs build` gates do not apply.

`edge` was merged in at f0851f6. Both sides had rewritten the commands table — #18 changed what `/rev` and `/adv-rev` say about posting to the thread, this branch added three rows — and the resolution keeps both. `check-commit-message.py --range` is clean over all four commits, `check-frontmatter.py` reports 12 documents carrying their frontmatter, and `epic.yml` parses.

## Not covered

- The existing five `tasks.md` files are untouched. They hold completed history, and migrating them to issues is separate work that should not ride on a change defining the stages. The `Task Index` in `internal/features/README.md` still links them, because they still exist.
- Nothing enforces any of this: a decomposition can still file an epic with a status, or nest two levels, and only a reviewer would notice. That is the subject of #13.
- The command files are still five pairs of byte-identical copies with nothing comparing them. #20 owns that, because the fix is a choice between a check and a stub that this change does not record.
- `decompose` does not say how much of a backlog to file at once beyond "an epic of twenty is a backlog wearing a stop rule". The right size is not known yet.


internal/thunderstorm.md
desertthunder: is this section necessary?

internal/features/README.md
```diff
@@ -1,14 +1,18 @@
...
-features/<feature-name>/{plan,tasks}.md
+features/<feature-name>/plan.md
 ```
 
-Checked tasks describe completed work. A completed track stays here until it
-ships; its durable decisions can then move to `../archive/`.
+A completed track stays here until it ships; its durable decisions can then
+move to `../archive/`.
+
+The tracks below also carry a `tasks.md`. Those predate the board and are not
```
desertthunder: the tracks below don't need to be mentioned here. they're going to go away soon.

desertthunder:
## Adversarial pass

Prose and skills only, so the hunt is for invariants the next stage relies on
and this one does not establish. Claims below were checked against the files on
`a5c57fc`, not recalled.

`high` · `.claude/skills/decompose/SKILL.md`:39 — the skill never says to create
the GitHub parent-child relation, only to "file the sub-issues under it" →
`/thunderstorm` reads children through the hierarchy API (`issue_read` method
`get_sub_issues`, `gh issue view --json subIssues`) and its start gate stops on
"It declares no sub-issues" (`.claude/skills/thunderstorm/SKILL.md`:31). File an
epic and four sub-issues listing them as the `- [ ]` lines `epic.yml` asks for,
label everything correctly, and `/thunderstorm <epic>` refuses to start: the
relation was never created. `github-board` has the call, but it lives under
**File new work** — work found mid-run — so a decomposition following this skill
lands on `issue_write` `create` with labels and no parent → say in **The
sub-issues** that the link is the API relation: `parent_issue_number` on create,
or `sub_issue_write` method `add` afterwards, which takes the sub-issue's **ID**
rather than its number. Worth saying the epic's checklist is not the link, for
the same reason this PR removes `tasks.md`: a `- [ ]` list beside a tracked
relation is the second copy of board state that is wrong the first time an issue
closes.

`medium` · `.claude/skills/specify/SKILL.md`:83 — "Hand it to `decompose`" sets
no order against the review the same file requires at :74 → a spec lives on
`agent/<name>` until it merges, and `decompose` cites its identifier at :51.
Decompose straight off the branch and the issues cite a ULID that resolves to
nothing on `edge`; a worker then runs `/impl`, whose worktree is cut from
`origin/edge` (`.claude/skills/implement/SKILL.md`:46) and whose first read is
"the spec the issue points at" (:31), and finds no file at that path. A review
round that changes the decision leaves issues cut from a document that no longer
says what they were cut from → state the order in **Then**: decomposition runs
after the spec merges to `edge`.

`medium` · `.claude/skills/decompose/SKILL.md`:33 — "Use
`.github/ISSUE_TEMPLATE/epic.yml`" is not executable through either transport
`github-board` defines, and the field list is short one required field → issue
*forms* render in the web UI only. Neither `issue_write` `create` nor `gh issue
create` applies the body or the `labels: ["kind:epic"]` line, so what an agent
actually does is reproduce the fields it was told about: "the goal, the stop
rule, the planned sub-issues, the source, and what the epic must not absorb".
That drops **Risk**, which is `required: true` (`epic.yml`:79), and the dropdown
produces body text rather than a `risk:*` label, while :41 requires one on every
sub-issue and says nothing about the epic → say the fields are reproduced as
headings because the form does not apply over the API, list all six, and settle
whether an epic carries `risk:*`.

`low` · `.claude/skills/decompose/SKILL.md`:34 — the stop-rule rule contradicts
the template it cites one line earlier → :35 forbids "a restatement of 'every
sub-issue is done'"; `epic.yml`:34 offers exactly that as its placeholder:
"Every sub-issue is status:done, and `cargo test --workspace` and `pnpm --dir
docs build` both pass on edge." An agent filling the template copies the
example, and the rule is only in a skill it may not have open → fix the
placeholder rather than the skill, so the example teaches the rule.

`low` · `internal/thunderstorm.md`:88 — the Stages block names "the whole path"
and omits `/thunderstorm` → it runs `/r-d` → `[/spec-ify]` → `/decomp` →
`/impl` → review passes → a human merges, with nothing between the epic and the
pull request. The run loop is what claims each sub-issue and drives those review
passes, it is the subject of the document the block sits in, and `epic.yml` tells
a filer the epic is "started by `/thunderstorm <number>`" → add the stage between
`/decomp` and `/impl`. The Commands table above it has the same gap, which
predates this change but is now the only table a reader has for the path this
section just promised.

`nit` · `.claude/commands/decomp.md`:1 — two more byte-identical alias pairs, and
nothing compares them → five pairs now (`r-d`/`rubber-duck`, `impl`/`implement`,
`edit`/`revise`, `decomp`/`decompose`, `spec-ify`/`specify`). All identical
today; an edit to one copy gives a repository where `/decomp` and `/decompose`
do different things under one skill name, and only a reader who opens both
notices. `internal/thunderstorm.md`:189 asks for a check rather than a rule to
remember → a loop in CI comparing each pair costs about three lines, or the
short name becomes a stub that names the long one.

Not a finding, recorded so the next pass does not re-derive it: the six other
claims these files make check out — `epic.yml`, `labels.yml`, and `ulid.py`
exist; the frontmatter triple at `specify`:55 matches
`internal/thunderstorm.md`:219; `implement`:26 does refuse an issue with
children as `decompose`:58 claims; and the branch rules already cover a spec
branch with no issue behind it.

No code in this diff, so nothing here is about tests. Nothing enforces any of
it either, which the pull request says and #13 owns.

— claude-opus-5 · high

---

desertthunder:
## Edit pass

Against the adversarial findings in [#issuecomment-5721370590](https://github.com/stormlightlabs/thunderus/pull/15#issuecomment-5721370590). Pushed as 06c548d.

Both the review and this pass ran in one session, so the separation the sequence exists to create is not there. Read the second column knowing that.

| Finding | Outcome |
| --- | --- |
| `high` decompose:39 — no parent-child relation | Fixed |
| `medium` specify:83 — no order against the spec's merge | Fixed |
| `medium` decompose:33 — template unusable over the API, Risk dropped | Fixed |
| `low` decompose:34 — placeholder contradicts the stop-rule rule | Fixed |
| `low` thunderstorm.md:88 — Stages omits `/thunderstorm` | Fixed |
| `nit` commands — alias pairs with nothing comparing them | Deferred, #20 |

**`high` · the relation.** **The sub-issues** now ends by saying the link is the relation, that a run asks GitHub for the epic's children and refuses an epic declaring none, and where `github-board` keeps the calls. The finding's other half was the cause: `epic.yml` asked for `- [ ]` beside a relation GitHub already tracks, which is the second copy of board state this pull request removes from `tasks.md` and had reintroduced in the epic. Those are now plain lines, and the field says the relation is what a run reads.

**`medium` · ordering.** `specify`'s **Then** hands off after the pull request merges, naming both costs: a worktree cut from `edge` has no such file, and review can still change the decisions the issues were cut from. `decompose`'s opening carries the same sentence, since `/decomp` can be pointed at a spec without `specify` running first.

**`medium` · the template.** Rewritten to say the form renders in the web UI and nowhere else, so filing means reproducing the six fields as headings and setting `kind:epic` by hand. The field list now includes Risk.

One decision here that the review asked for and no document records: **an epic takes no `risk:*` label**, and risk stays in the body where the dropdown puts it. The reasoning is that the label sizes the blast radius of a change and an epic makes none — the same reason an epic carries no status. It is written down in `decompose` rather than left implied, so overruling it is one edit. Say so if the epic should carry the label instead.

**`low` · the placeholder.** Replaced with a stop rule that is the epic's own. The skill's rule was correct, so the example was what needed fixing.

**`low` · the Stages block.** `/thunderstorm` sits between `/decomp` and `/impl` in both the block and the Commands table, with a line saying a run claims each sub-issue and dispatches the stages under it, and that a single issue skips the run. No command file was added: the skill is invocable by name, and a twelfth file in `.claude/commands/` would extend the duplication the `nit` is about.

**`nit` · deferred to #20.** Real, and this pull request added two of the five pairs. It is deferred rather than fixed because the fix is a choice — a derived check in CI, or the short name becoming a stub so there is one copy and nothing to compare — and the second changes what a reader of `/r-d` sees. That choice is not recorded anywhere, and a CI change does not belong in a pull request defining stages. The issue states both shapes and the criteria either must meet.

No tests: the diff is prose, skill files, and an issue template. The `high` finding therefore has no regression test, and nothing here could have one until #13 exists — which is the same gap the pull request already lists under **Not covered**.

Verification: `check-commit-message.py` clean on both commits, `epic.yml` parses, remote `agent/spec-and-decomp` confirmed at 06c548d. No Rust or `docs/` changed, so the cargo and `pnpm --dir docs build` gates do not apply.

— claude-opus-5 · high

---

internal/thunderstorm.md
desertthunder: No, and it is gone in d47fa64. A heading whose content is the absence of a list reads as a placeholder, and the section was doing two unrelated jobs anyway.

Both rules moved to **File conventions**, which is what they were about:

> An issue cut from a document cites that document's identifier — this one is `01M2PWX233GKXE5M9SPTNTGN0D` — so the trail is readable from either end. What no document carries is a list of pending work: the board holds that, and a list of gaps goes stale the moment one closes. Two bullets in the list this document replaced were already false within hours of being written.

The identifier line had to survive somewhere: `decompose` tells a filer to cite the source document, and this document is a source. The staleness rule is the same lesson **Recording a failure mode** keeps, so it is worth a sentence, just not a section.

Leaving this thread for you since it was a question and dropping the section was my call.

— claude-opus-5 · high

---

internal/features/README.md
desertthunder: Removed in d47fa64. The paragraph is down to the one line that is not about the legacy tracks:

> Write the plan with `/spec-ify`, then file the work with `/decomp`.

One copy of that note is still in `specify`, under **What a spec does not hold**, because a writer who opens `internal/features/` sees five directories that each carry a `tasks.md` and will copy the pattern unless something says not to. That one is aimed at the writer rather than at a reader of the index, so it earns its place until the files are actually gone — say the word and it goes now.

— claude-opus-5 · high

---

desertthunder:
## Edit pass

Conflict resolved and both review threads addressed. Head is d47fa64.

| Item | Outcome |
| --- | --- |
| Merge conflict against `edge` | Resolved, f0851f6 |
| `internal/thunderstorm.md`:274 — is `What is not done yet` necessary? | Fixed, [thread](https://github.com/stormlightlabs/thunderus/pull/15#discussion_r4041647205) |
| `internal/features/README.md`:12 — the legacy tracks do not belong here | Fixed, [thread](https://github.com/stormlightlabs/thunderus/pull/15#discussion_r4041647766) |

**The conflict.** `edge` had moved three commits and both sides rewrote the commands table: #18 changed what `/rev` and `/adv-rev` say about posting to the thread, this branch added rows for `/spec-ify`, `/decomp`, and `/thunderstorm`. The resolution keeps both, so `/rev` still reads "Comments only on the second pass" and `/adv-rev` "Always comments". Nothing else conflicted; #10's frontmatter and #11's worktree changes merged clean.

Merged rather than rebased, because the branch is published and someone may have it checked out. The merge commit is `chore:` rather than git's default subject — `check-commit-message.py` rejects `Merge branch ...`, and a message is only fixable before it is pushed.

**`What is not done yet`.** Dropped. Its two rules moved into **File conventions**, which is what they were about: an issue cites the identifier of the document it came from, and no document carries a backlog because the board does. Left that thread open, since it was a question and the call to delete was mine.

**The features README.** Down to `Write the plan with /spec-ify, then file the work with /decomp.` The note about the legacy `tasks.md` files survives only in `specify`, aimed at the writer who sees five sibling directories and would copy the pattern. Thread resolved; the detail is on it.

Two claims in the pull request body went stale and are updated: it said the README explains the legacy tracks, and it counted two commits.

Verification: `check-commit-message.py --range origin/edge..HEAD` clean over all four commits, `check-frontmatter.py` reports 12 documents carrying their frontmatter, `epic.yml` parses, remote `agent/spec-and-decomp` confirmed at d47fa64. Still no Rust or `docs/`, so the cargo and `pnpm --dir docs build` gates do not apply.

Standing: #20 is the one deferred item, and nothing from the adversarial pass is outstanding.

— claude-opus-5 · high

---

PR #18, opened 2026-09-17
feat: say which review passes post to the thread

desertthunder:
## What

Writes down which review passes comment on a pull request. A first standard pass reports to its invoker only, a second comments on what survived it, and an adversarial pass always comments. `review/SKILL.md` gains the rule, and `rev.md` and `adv-rev.md` stop contradicting it.

## Why

The rule was agreed and never recorded. `rev.md` still said "Post one comment with all findings", so a reviewer following the files and a reviewer following the policy would behave differently, and the files win by default. That happened on #13 this afternoon: the first pass held its findings back because the policy said so, while the command said to post.

A first pass mostly finds things that are fixed within the hour. A thread of resolved findings costs a later reader attention and teaches them nothing. What survives a round of fixes is the part worth recording, which is why the second pass comments and the first does not.

An adversarial pass always comments, including to say it found nothing, because it is the last pass before a human reads the diff and the record has to show it ran.

Which pass is which comes from the pull request rather than from anyone's memory: no signed review findings on it means this is the first.

Both commands also gained `complexity`, which `review/SKILL.md` has covered since #2 but neither command mentioned.

## Verification

Read against `review/SKILL.md` as it stands, which already required one comment per pass and a model signature. The new section sits before the posting mechanics rather than replacing them, so transports and the signature format are unchanged.

Prose only.

## Not covered

- Nothing enforces this. A pass that comments when it should not is caught by a reader, or not at all.
- The rule assumes signed findings are how you recognise a previous pass. A human review without a signature does not count, which is probably right, but it is an assumption rather than a check.
- `revise` says who resolves a thread; this says who opens one. Neither says what happens when a pass finds something a previous pass had already resolved.

---


desertthunder:
Adversarial pass. The change reads coherently on its own; the defects are in what executes it. Every reference below was checked against the post-change files.

**blocker** · `.claude/skills/review/SKILL.md:98` — The pass-number test is a fixed point that never advances. Pass 2 is distinguished by signed review findings existing on the pull request, but pass 1 is told to leave none, so at pass 2 the thread still holds no review findings and pass 2 also classifies itself as first. Under the rule as written no standard pass ever comments, and the thread ends up holding only `/edit` replies plus the adversarial comment — losing the standard-review record this change exists to regulate. → Have the invoker state the pass; `.claude/skills/thunderstorm/SKILL.md:60` already fixes the sequence at dispatch.

**blocker** · `.claude/skills/revise/SKILL.md:19` — `/edit`'s only documented input is the pull request thread (`gh pr view <n> --comments`, MCP `get_comments`/`get_review_comments`, then SKILL.md:26 "Collect every finding from the most recent review pass"). Pass 1 now posts nothing, so step 2 of the sequence at `internal/thunderstorm.md:104` has no input at all. The findings live only in the orchestrator's transient context, and `.claude/agents/reviewer.md` is a subagent that returns to its invoker with no instruction to relay them anywhere. → Give `/edit` a documented non-thread channel for first-pass findings, or keep pass 1 posting.

**high** · `internal/thunderstorm.md:103` — Still states the opposite of the new rule: line 103 "1. `/rev` posts findings as pull request comments", and line 94 "`/rev` … Standard review pass. Posts findings as comments." The stated purpose of this change is to say which passes post, and the authoritative workflow document still says all of them do, so an agent that reads `internal/thunderstorm.md` first executes the old contract. → Update lines 94 and 103 in the same commit, or the rule is advisory only.

**high** · `.claude/skills/review/SKILL.md:101` — "only on what survived the first" is not computable by the pass asked to compute it. `.claude/agents/reviewer.md` runs as a fresh subagent per pass with no memory of pass 1, and pass 1's findings are now written to no durable surface; on a run resumed in a new session they are gone outright. Pass 2 either re-reports everything it independently finds, defeating the filter, or silently under-reports survivors it cannot recognise. → Record pass 1's findings somewhere durable, or drop the "survived" filter and let `/edit` outcomes mark what closed.

**high** · `.claude/skills/review/SKILL.md:107` — "Findings go to whoever invoked the pass" is the only remaining destination for a first pass, and when the pass runs as a subagent the invoker is another agent, not a human. The line deleted at SKILL.md:124 ("Print the same findings in chat") was the one sentence guaranteeing a human-visible surface. Combined with pass 1 not posting, a first pass's findings can terminate in an agent context that is discarded, with no file claiming otherwise. → Restore an explicit human-reachable destination for a non-posting pass.

**medium** · `internal/thunderstorm.md:67` — `status:review` "leaves when all three review passes clear", but under the new rule passes 1 and 2 may leave no trace on the pull request. No observer, no human merging, and no resumed run can tell whether three passes ran; a re-run of pass 1 is indistinguishable from a first run, so it re-classifies itself as first indefinitely. `internal/models.md:38` (the two standard passes should not share a model and level) also becomes unverifiable from the record. → Have every pass leave a signed marker even when it posts no findings, which is what the adversarial pass is already told to do.

**medium** · `.claude/skills/review/SKILL.md:98` — Undefined inputs to the detection rule, each silently changing which pass the agent believes it is: no pull request at all (SKILL.md:12 explicitly allows a branch name or an empty target, and SKILL.md:124 covers only the posting question, not the detection one); a human review carrying no signature, which leaves a human-reviewed pull request reading as first; findings posted and then deleted or edited away; a force-push that rewrites the diff a "first" pass already reviewed. → State the behaviour for each, or remove the dependency on thread state.

**medium** · `.claude/skills/review/SKILL.md:98` — False positives, and no locking. `.claude/skills/revise/SKILL.md:77` makes every `/edit` reply signed and finding-shaped ("listing each finding and its outcome"), and nothing distinguishes a reviewer's signature from an editor's, so a genuine first pass on a pull request that already had an editor reply reads it as prior review findings and posts. Note this interlocks with the first blocker: the only thing that breaks the fixed point is this misreading, so the sequence currently works, when it works, by accident. The thread is also now the pass-identity lock with no locking — two concurrent passes both read "first", and a pass whose comment fails to post silently demotes the next one. → Mark reviewer comments with a distinguishable token, or carry the pass number out of band.

**medium** · `.claude/commands/rev.md:12` — Two different tests for one condition, in files changed by the same commit: `rev.md` says post "only if a signed review pass already ran against it", SKILL.md:98 says "no signed review findings on it means this is the first". A pass that ran and found nothing satisfies the first and fails the second. `rev.md:13` ("in which case report only what survived") also reads against `rev.md:11` ("Report every finding to me") and against SKILL.md:107, which says findings go to the invoker either way. → Settle on one test and one reporting rule, stated once.

**medium** · `.claude/skills/review/SKILL.md:91` — Complexity. Deciding the pass number by archaeology on the pull request thread is harder than the problem it solves: it adds stateful inference over a mutable, externally editable surface, with at least four undefined inputs, no idempotence and no concurrency story, to decide one boolean the caller already knows. The simpler version is that the invoker names the pass, since `.claude/skills/thunderstorm/SKILL.md:60` fixes the sequence at dispatch and `/rev` already takes an argument. Cost: a human running `/rev` by hand has to say which pass, or gets the safe default of report-to-invoker and do not post. That one change also removes the first blocker and three of the findings above it, which all stem from the inference.

**low** · `.claude/commands/adv-rev.md:12` — "Do not repeat earlier findings", and SKILL.md:130 "Do not repeat a finding already posted and addressed in an earlier pass", are unenforceable for the last pass once the standard passes post nothing: the only earlier findings it can observe are none. The adversarial pass will re-raise standard findings it cannot see were already raised and fixed — the duplication the rule forbids. → The same durable record fixes this.

**low** · `.claude/skills/review/SKILL.md:91` — "Verification: prose only" understates the gap. The only automated check over `.claude/` is `.claude/scripts/check-frontmatter.py`, which validates frontmatter and passes regardless of whether the protocol these files define is coherent, so the three contradicting documents above ship green. → If the protocol is going to span seven files, a consistency check, or consolidation into one, is what keeps it true.

— claude-opus-5 · high

---

desertthunder:
Edit pass answering the adversarial pass above (`claude-opus-5 · high`). Pushed as `544d3a4` and `9554312`.

Eleven fixed, one deferred. Both blockers, the complexity finding and four of the mediums shared one cause — deriving the pass number from the thread — so they are one change: the invoker names the pass, `/rev` takes it as an argument, and an unnamed pass is a first pass, which is the safe end.

| # | Finding | Outcome |
| - | ------- | ------- |
| 1 | `blocker` pass-number fixed point | Fixed |
| 2 | `blocker` `/edit` left with no input | Fixed |
| 3 | `high` `internal/thunderstorm.md` contradicts the rule | Fixed |
| 4 | `high` "what survived" not computable | Fixed |
| 5 | `high` no human-reachable destination | Fixed |
| 6 | `medium` `status:review` unobservable | Fixed |
| 7 | `medium` undefined inputs to detection | Fixed |
| 8 | `medium` false positives, no locking | Fixed |
| 9 | `medium` `rev.md` and SKILL.md disagree | Fixed |
| 10 | `medium` complexity of thread archaeology | Fixed |
| 11 | `low` "do not repeat earlier findings" unenforceable | Fixed |
| 12 | `low` nothing checks the protocol | Deferred — #22 |

**1, 7, 8, 10 — Fixed.** The inference is gone. `review/SKILL.md` now says the invoker says which pass this is, `/rev` takes `first` or `second` as an argument, and `thunderstorm/SKILL.md` says to tell each pass which one it is at dispatch. Where no pass is named it is a first pass: report to the invoker, post nothing. That removes the fixed point, the four undefined inputs, the `/edit`-reply false positive and the unlocked shared state in one move, because none of them exist once nothing reads the thread to decide. The three ways the thread answers wrongly are written down so the next person does not reintroduce it.

**2 — Fixed.** `revise/SKILL.md` now says a first pass's findings reach `/edit` from whoever invoked it rather than from the thread, and to read the thread as well since later passes post there. Where nothing was handed over and the thread is empty, `/edit` says so and stops instead of inventing a pass to address. `thunderstorm/SKILL.md` says the orchestrator hands them over, so the claim has a holder.

**3 — Fixed.** Both places: the command table rows for `/rev` and `/adv-rev`, and steps 1–5 of the review sequence.

**4 — Fixed, and it was also wrong in my first commit.** `544d3a4` said the second pass posts what survived without saying how it learns what the first found — the same defect, one layer down. `9554312` adds it: the orchestrator hands the first pass's findings to the second pass as well as to `/edit`.

**5 — Fixed.** "Print the findings in chat" is restored to the posting section, explicitly including the case where nothing is posted.

**6 — Fixed.** Every `/edit` reply now names the pass it answers and the signature that pass ran under, so a first pass leaves its trace in the reply to it. The thread reads `/edit` reply, second-pass comment, `/edit` reply, adversarial comment, `/edit` reply — all three passes legible without any of them posting a bare marker. This also answers the `internal/models.md:38` half: the reply carries the model and level the pass ran at, so the two-standard-passes rule stays checkable after the fact.

**9 — Fixed.** `rev.md` no longer carries its own test. It states the same rule as `review/SKILL.md` and nothing else.

**11 — Fixed.** `adv-rev.md` and the stop condition now say not to repeat a finding an earlier pass raised "whether it is on the thread or was handed to you", which is a condition the pass can actually evaluate.

**12 — Deferred to #22.** Real, and out of scope for a change this size: closing it means either a machine-readable table the prose is checked against, or consolidating seven files into one.

One correction to my own review, since it was wrong on the record: finding 12 cited `.claude/scripts/check-frontmatter.py` as the only automated check over `.claude/`. That script is not on this branch — it arrived on `edge` in `f51217e`, which is ahead of this pull request's base, and I read it off another checkout. On this branch the only thing CI runs over `.claude/` is `check-commit-message.py`. The finding stands and is slightly stronger than stated; the file named in it does not.

Not run: `cargo fmt`, `clippy` and `cargo test`. No Rust changed — the diff is six Markdown files — and `CLAUDE.md` scopes those to Rust changes. Nothing under `docs/` changed, so no docs build.

Threads: none to resolve. The findings were posted as one comment rather than line threads, and it is my own review, which I do not get to mark closed.

— claude-opus-5 · high

---

Issue #26, opened 2026-09-17
Verify the interface with rendered terminal captures


desertthunder:
From `01M2RNSY07A522766KBV2DPR41`, `internal/features/tui-verification/plan.md`, which answers the three questions left open by the idea `01M2RH1Z3FCE0WBY5H6G9VAGS0` in `internal/ideas/tui-screenshot-verification.md`.

Reshaped after the original design was found to rest on a false premise. The frames are rendered and looked at rather than stripped to text and posted as a pull request comment. [What was dropped and why](https://github.com/stormlightlabs/thunderus/issues/26#issuecomment-5722885428).

## Goal

A change to the interface produces frames someone looks at. Twelve named scenarios are captured in one pass from generated fixtures, and any of them renders to an image, so a reviewer sees the frame instead of inferring it from the code that drew it.

Nothing is captured outside the set, so a change cannot choose the frame that flatters it.

Nothing is committed. A frame is evidence read while the change is in front of someone, with no life after it lands.

## Done when

A capture run produces all twelve scenarios with the credential store empty and no provider environment variable set.

`freeze` renders a captured `.ansi` to an SVG that keeps its color.

`git status --porcelain` after a capture run reports nothing: the `.ansi` files land in ignored scratch space and no capture, image, or fixture is committed.

A `Theme` variant with no matching scenario fails `cargo test --workspace`.

`docs/src/content/docs/docs/development/tui-qa.md` teaches that harness and contains no fixed sleep and no `capture-pane` without `-S`.

## Why rendered, and why twelve

`freeze` reads a `capture-pane -e` file directly and keeps its escapes as color. Structure, spacing, alignment, truncation, and rhythm are what a capture is read for, and the palette comes along instead of being thrown away.

That is why all twelve scenarios stay in the set. The earlier shape of this epic posted nine, cutting the three `theme-*` scenarios and `no-color` because stripped of escapes all four are the same text as each other. Rendered they are not, so each carries something a reader can act on, and the check in #33 against the `Theme` enum has a subject again.

Render to SVG. `freeze` v0.2.2 segfaults rasterizing PNG in the cloud container; the crash reproduces on plain-text input with no ANSI at all, so it is the rasterizer rather than the capture. SVG succeeds on the same input.

The whole route was run end to end before this body was rewritten: capture at 100x30, 60x30, and 100x16, poll until the pane settles rather than sleeping, render each to SVG, twelve distinct fill colors surviving, and an empty `git status --porcelain` after.

[...]

PR #47, opened 2026-09-17
test: assert fake-agent clears first-run setup

### LINKED ISSUE #28: Reach the transcript surface without a provider credential
From `01M2RNSY07A522766KBV2DPR41`, `internal/features/tui-verification/plan.md`. Nothing blocks this. The capture harness in this epic waits on it.

## Problem

A capture never sends a request and the container holds no key, so the harness has to reach the main surface with the credential store empty. It cannot today.

First-run setup opens whenever the selected provider has no credential (`crates/thndrs/src/cli/app/onboarding.rs`), and `SetupProviderArg` (`crates/thndrs/src/cli/commands/setup.rs:39`) offers four routes: `Umans`, `OpencodeGo`, `OpencodeZen`, and `ChatgptCodex`. The first is retired and kept only so old command and config values produce the same unsupported-route diagnostic. The other three want a credential. A capture run in a clean container therefore renders onboarding, and every scenario past that is unreachable.

The spec makes this the first question of the track: establish which existing route clears the gate, and if none does, adding one is in scope.

## Done when

A command sequence starts `thndrs` with no provider environment variable set and an empty credential store, and the process renders the transcript surface rather than the onboarding surface. A test asserts that under the same empty environment.

If the sequence needs a new route, that route neither accepts nor writes a credential that a real run would then use, and a test asserts it: a run through the route leaves the credential store empty and sends no provider request.

The sequence is recorded in `internal/features/tui-verification/plan.md`, so the harness issue can cite it rather than rediscover it.

## Not in this issue

The tmux harness, the fixtures, and the captures themselves. This issue establishes only that the surface is reachable.

Changing what onboarding does when a credential is genuinely wanted. The recovery stages in `onboarding.rs` are unchanged.

desertthunder:
Closes #28

## What

`--model fake-agent` is the route that reaches the transcript surface with an
empty credential store. No application code changes: `selected_provider_missing`
already returns `None` for the prefix (`crates/thndrs/src/cli/app/onboarding.rs:425`)
and `model_uses_unsupported_route` already exempts it
(`crates/thndrs/src/cli/commands/setup.rs:30`). Three tests hold that behavior
against a later edit, and `internal/features/tui-verification/plan.md` records
the launch line so the harness issue can cite it.

## Why

A capture never sends a request and the container holds no key, so every
scenario past `startup` was unreachable while onboarding opened on launch. The
first question of the track was which existing route clears the gate; adding one
was in scope only if none did, and one does.

## Verification

- `cargo fmt`, then `cargo clippy --workspace --all-targets --all-features --locked -- -D warnings` — clean.
- `cargo test --workspace --all-features --locked` — 1849 passed, 0 failed, 12 ignored.
- End-to-end under tmux, `env -i` with `HOME` on an empty directory: the recorded
  launch line drew the composer and `✓ Ready … fake-agent · auto`, and the same
  launch without `--model` drew `setup · chatgpt-codex · choose provider`.
  `/resume` reached the session lookup rather than the ephemeral refusal.
- Both route tests were mutation-checked: swapping the model to
  `opencode/big-pickle` or to an unsupported id fails each of them, and the
  seeded-credential run emits `RequestStarted` followed by a 401, which is what
  makes the dispatch assertion able to fail.
- No `docs/` change, so `pnpm --dir docs build` was not run. Only `internal/`
  prose changed, which `CLAUDE.md` exempts.

## Not covered

The tmux harness, the fixture generator, and the captures themselves stay with
the other issues under #26. Nothing in `onboarding.rs` changes, so the recovery
stages for a credential that is genuinely wanted behave as before. No test was
changed, narrowed, or removed. `docs/src/content/docs/docs/development/tui-qa.md`
still teaches the sleep-and-visible-pane pattern; the plan assigns its rewrite to
the harness work, not here. Every review pass on this branch so far was written
by the session that wrote the diff.

https://claude.ai/code/session_016WjCmEDAmjmAj7TXTYCxFn


desertthunder:
Edit pass answering the first standard review, which read `84a0ac2` and posted nothing to this thread. Both the review and this edit ran on a diff written in the same session that wrote the pull request. Now at `8c63632`.

| Finding | Outcome |
| --- | --- |
| `high` · plan launch line carries `--ephemeral`, which `resume_session` rejects | Fixed in `eba345e` |
| `medium` · credential-store test never ran a turn, so it held for any model | Fixed in `8c63632` |
| `low` · parse test bypassed the check it named, via `--model` | Fixed in `8c63632` |
| `low` · accessory assertion could not tell the two states apart | Fixed in `8c63632` |
| `nit` · `ProviderKind` row misplaced, doc comment overclaimed | Fixed in `8c63632` |
| `nit` · plan paragraph restated what the tests assert | Fixed in `eba345e` |

The `high` row was the one that mattered: `/resume` fails with `cannot resume a session in ephemeral mode` (`crates/thndrs/src/cli/app.rs:1613`), so the recorded line reached only `startup` and would have misdirected the harness issue. It now uses a scratch `--session-dir` and no `--ephemeral`. Confirmed under tmux: `/resume nonexistent-session` returns `session ... is not found`, not the ephemeral refusal, and the session file lands in the scratch directory.

The `medium` row is fixed at the cause rather than by renaming the test. It drives a whole turn through `RunHandle::provider_with_steering`, which picks the provider from the model instead of being handed one, and asserts no `AgentEvent::RequestStarted` was dispatched. Two mutation checks confirm it now bites: `big-pickle` (unsupported) and `opencode/big-pickle` (real provider, no credential) each fail it. `RunHandle::fake` was unsuitable here because it hardcodes `ProviderKind::Fake` and so cannot discriminate on the model.

No test was weakened, skipped, or narrowed. Verification: `cargo fmt`, `cargo clippy --workspace --all-targets --all-features --locked -- -D warnings` clean, `cargo test --workspace --all-features --locked` 1849 passed / 0 failed / 12 ignored. The lib count drops by one against `84a0ac2` because three tests became two and the `ProviderKind` assertion folded into an existing case.

— claude-opus-5 · standard

---

desertthunder:
Adversarial pass. Reviewed at `8c63632` on #47. I wrote this diff and the two earlier passes on it in the same session, so this is not an independent reading.

`medium` · `crates/thndrs/src/cli/app/tests/setup.rs:854` — the `RequestStarted` assertion cannot fail under the conditions the test itself establishes. `with_setup_home` empties the credential store, and `run_provider` returns at `crates/thndrs/src/core/agent/run.rs:113` when `from_env_or_dotenv` cannot build a client, so no real provider ever reaches `crates/thndrs/src/core/agent/execution.rs:216`, the only site that emits the event. The mutation run behind the last commit confirms it: `opencode/big-pickle` fails on the `Finished` assertion, never on this one, which also makes the doc comment at `setup.rs:828` name a failure mode that does not occur. → Seed a credential into the project store before the turn, so a real provider reaches dispatch and the assertion goes live. That also proves the stronger property the issue asks for: the route dispatches nothing even when a credential is sitting there. Keep the store-is-empty assertions in the other test, which has no credential to seed.

`medium` · `internal/features/tui-verification/plan.md:66` — "`--session-dir` is global (`crates/thndrs/src/cli/mod.rs:278`)" is false. That field carries only `#[arg(long = "session-dir")]`, with no `global = true`, unlike `ephemeral` two lines below it. `thndrs run --session-dir /tmp/x hi` answers `error: unexpected argument '--session-dir' found`. The text predates this pull request, but the new launch line at `plan.md:100` now puts that flag in every capture invocation, so the document has become load-bearing on a claim the code does not support. It works today only because the line is a top-level invocation with no subcommand. → Correct the sentence in this pull request rather than leaving the harness issue to inherit it.

`low` · `crates/thndrs/src/cli/app/tests/setup.rs:836` — `App::from_cli` is decorative in the turn test. `app.runtime.model` is `Cli.model` copied verbatim (`crates/thndrs/src/cli/app.rs:1539`) and no assertion afterwards touches the app, so the test has the same power with the literal in its place. It reads as app-level coverage that is not there. → Drop it, or assert something about the app that the turn changed.

`nit` · `crates/thndrs/src/cli/app/tests/setup.rs:817` — both assertions land on the same early return: `onboarding.rs:425` returns before `pending_provider_prompt` is read, so passing `true` here exercises nothing the build-time `false` did not.

No races, no unhandled input, and no security findings. The turn is fully drained inside the env guard, so the worker cannot outlive the restored environment.

— claude-opus-5 · adversarial

---

desertthunder:
Edit pass answering the adversarial review at `8c63632`, which ran as `claude-opus-5 · adversarial`. Now at `14542d0`. As noted there, the same session wrote the diff and all three passes over it.

| Finding | Outcome |
| --- | --- |
| `medium` · `RequestStarted` assertion could not fail | Fixed in `14542d0` |
| `medium` · plan calls `--session-dir` global | Fixed in `4020f67` |
| `low` · `App::from_cli` decorative in the turn test | Fixed in `14542d0` |
| `nit` · two assertions on one early return | Fixed in `14542d0` |

The first row was right, and the fix confirms it from the event stream rather than by argument. With the store empty, `run_provider` gives up at `crates/thndrs/src/core/agent/run.rs:113` and returns `Failed`, so nothing ever reached `execution.rs:216`. The turn test now seeds a credential before running, which lets a real provider build its client and dispatch. Under `opencode/big-pickle` the stream carries `RequestStarted` followed by `Failed("authentication failed (HTTP 401)")`, and the test stops on `the route must dispatch no provider request even with a credential present` — the assertion named in the finding, by name. The assertions were also reordered so the dispatch check runs before the completion check; previously the stricter `Finished` assertion tripped first and hid which property had actually broken.

Seeding is also the stronger property: the route dispatches nothing *while a usable credential sits in the store*, and the store is byte-identical afterward. The empty-store assertions moved to the startup test, which is where the store is genuinely empty.

On the second row: `thndrs run --session-dir /tmp/x hi` answers `error: unexpected argument '--session-dir' found`, so the sentence is now corrected to say top-level rather than global. The launch line already placed the flag correctly.

No test was weakened, skipped, or narrowed, and nothing was deferred. `cargo fmt` and `cargo clippy --workspace --all-targets --all-features --locked -- -D warnings` are clean; `cargo test --workspace --all-features --locked` gives 1849 passed, 0 failed, 12 ignored. The mutation runs above were reverted before committing; the suite itself never dispatches a request, so it makes no network call.

— claude-opus-5 · standard

---

desertthunder:
Adversarial pass. Reviewed at `86efcb5` on #47, after the merge of `ae8c53b`. Same session wrote the diff and every pass over it, including this one.

`medium` · `crates/thndrs/src/cli/app/tests/setup.rs:811` and `:858` — both tests pass one tempdir as HOME and the same tempdir as the workspace, so the two credential stores are one file. `global_credentials_path` is `home_dir().join(".thndrs").join("credentials.env")` (`crates/thndrs/src/core/auth.rs:350`), `home_dir` is `env::var_os("HOME")` (`crates/thndrs/src/core/utils/mod.rs:29`), and `project_credentials_path(&cwd)` is `cwd.join(".thndrs").join("credentials.env")` (`auth.rs:628`). With HOME equal to cwd those resolve to the same path, so the assertions at `setup.rs:822` and `setup.rs:826` check one file twice, and the credential the turn test seeds into what it calls the project store is simultaneously the global store. A regression that wrote to one store and not the other passes both tests unchanged. It also hands `App::from_cli` a workspace rooted at the home directory, which no real run has. Traced through the path construction rather than reproduced. → Take two tempdirs, one for HOME and one for cwd. Until then `plan.md:141` claims more than the test shows when it says every credential store is absent.

`low` · `crates/thndrs/src/cli/mod.rs:1379` — this is the only one of the three new tests that does not guard HOME, and `load_effective` reads the real `~/.thndrs/config.toml` (`crates/thndrs/src/core/config/mod.rs:633`) before the project layer. The `model` assertion is safe because project config merges over global, but a developer whose global config is not valid TOML gets `load_file` returning `Err`, which makes `.expect("load config")` panic on machine state the test never set. The two sibling tests in that module already have the same gap, so this follows the local convention rather than breaking it. → Wrap it in `with_home` over a temp directory.

`nit` · `internal/features/tui-verification/plan.md:98` and `:130` — the same directory is "a fixture directory" in Determinism and `<scratch>` in the launch line thirty lines later. One name would save a harness author the inference.

The merge itself is clean: the only delta between `origin/edge` and this head in `plan.md` is the two intended edits, and edge's rewritten Determinism opening survived intact. No races — the turn is drained inside the env guard, so the worker cannot outlive the restored environment. No security findings.

— claude-opus-5 · adversarial

---

desertthunder:
Edit pass answering the adversarial review at `86efcb5`, which ran as `claude-opus-5 · adversarial`. Now at `7ef80c7`.

| Finding | Outcome |
| --- | --- |
| `medium` · HOME and workspace were one directory, so both credential stores were one file | Fixed in `7ef80c7` |
| `low` · parse test read the real global config | Fixed in `7ef80c7` |
| `nit` · "fixture directory" and `<scratch>` named the same path | Fixed in `7ef80c7` |

The first row was right and is now demonstrated rather than argued. Each test takes two temporary directories, one for HOME and one for the workspace, so `global_credentials_path` and `project_credentials_path(&cwd)` are distinct files. The turn test gained an assertion that the global store stays absent after seeding the project store, which is only meaningful once they are separate — reverting just the split makes it fail with `the route must not copy the credential into the global store`, which is the proof that the two paths had been one.

For the second, `crate::test_env` now owns `with_home`, so `cli::tests` can reach it and the helper in `cli/app/tests/helpers.rs` delegates instead of repeating the unsafe environment dance. `load_effective` reads the global config before the project layer (`crates/thndrs/src/core/config/mod.rs:633`), so an unparseable one belonging to whoever ran the test used to fail the load; it now reads an empty temporary home.

Nothing was deferred, and no test was weakened, skipped, or narrowed. `cargo fmt` and `cargo clippy --workspace --all-targets --all-features --locked -- -D warnings` are clean; `cargo test --workspace --all-features --locked` gives 1849 passed, 0 failed, 12 ignored.

One honest note on that run: the first full pass after these edits failed `fake_client_smokes_cancellation` in `acp_server_smoke`, and a re-run passed. Everything in this commit is `cfg(test)` code inside the library or prose, none of which compiles into that integration binary, so the diff cannot reach it; it is a timing flake in a cancellation smoke test, not something this pull request introduced. Recording it rather than quietly reporting only the green run.

— claude-opus-5 · standard

---

PR #48, opened 2026-09-17
chore: install freeze from the session-start hook

### LINKED ISSUE #30: Install freeze from the session-start hook at a pinned version
From `01M2RNSY07A522766KBV2DPR41`, `internal/features/tui-verification/plan.md`. Nothing blocks this.

The acceptance criteria below are unchanged by the epic's move from committed captures to captures posted on a pull request. Only the framing in this section changed: `freeze` renders a local convenience rather than a step in the loop.

## Problem

A capture run writes its `.ansi` file to scratch space and posts the frame to the pull request as plain text. `freeze` is what turns that scratch file into a picture for whoever wants to look at one on the machine that produced it. A cloud container has no `freeze`, so that option is missing exactly where captures are taken.

`.claude/hooks/session-start.sh` already warms the Cargo registry and the docs dependencies in a report-and-continue style, reporting a failure to stderr and continuing rather than exiting non-zero (`:40`), and it exits immediately outside the cloud. Installing `freeze` belongs in the same place under the same rule, because a renderer that is only ever a convenience must never fail a capture run.

## Done when

`.claude/hooks/session-start.sh` installs `freeze` at a version pinned in the script. A failed install writes a message to stderr and the hook continues; the hook's exit status is unchanged by that failure. The hook still exits immediately outside the cloud.

The pinned version is v0.2.2 or later. A comment beside the pin records which attributes that version drops — `\e[3m` italic, `\e[2m` dim, `\e[7m` reverse, and basic backgrounds — and which it renders: bold, underline, every foreground, 256-color backgrounds, and truecolor. `ratatui_style` sets `ITALIC` and `DIM` (`crates/thndrs/src/cli/renderer/ratatui.rs:77-93`), so a regression in either leaves a rendered image unchanged, and a later reader needs to know that a frame correct in the `.ansi` file and wrong in the image is a `freeze` defect rather than an application one.

## Not in this issue

The capture harness's behavior when `freeze` is absent. The harness owns that, and a missing `freeze` does not stop a capture or a post.

Committing any image. None is committed, here or anywhere in this epic.

Changing the application to suit the renderer. The attributes `freeze` drops stay dropped.

desertthunder:
Closes #30

## What

`.claude/hooks/session-start.sh` installs `freeze` at `v0.2.2`, pinned in a
variable beside a comment recording which attributes that version drops (`\e[3m`
italic, `\e[2m` dim, `\e[7m` reverse, basic backgrounds) and which it renders
(bold, underline, every foreground, 256-color backgrounds, truecolor). The
install follows the report-and-continue rule the hook already uses: a failure
writes to stderr and the hook's exit status stays 0. The early exit outside the
cloud is unchanged. `GOBIN` points the binary at a directory on `PATH`, because
`go install` otherwise writes to `$(go env GOPATH)/bin`, which this image does
not carry on `PATH`, and `freeze --version` is then compared against the pin, so
a failed install cannot hide behind an older binary already on `PATH`.

`.claude/hooks/session-start-test.py` runs the hook against stub `rustup`,
`cargo`, `pnpm`, and `go` commands and holds that rule in place; a `hooks` job
runs it with no toolchain and no network.

## Why

A container carries no renderer, so a capture run there produces `.ansi` files
and nothing anyone can render. `internal/features/tui-verification/plan.md:189`
puts the install in this hook at a pinned version, and the comment on the pin is
what tells a later reader that a frame correct in the `.ansi` file and wrong in
the image is a `freeze` defect rather than an application one: `ratatui_style`
sets `ITALIC` and `DIM` (`crates/thndrs/src/cli/renderer/ratatui.rs:77-93`), and
`freeze` drops both, so a regression in either leaves the image unchanged.

Rebased onto `edge` at ae8c53b, which moved the capture set out of the tree and
into a pull request comment. The hook's comment and `internal/thunderstorm.md`
describe that workflow rather than the committed-capture one they were first
written against.

## Verification

- `.claude/hooks/session-start-test.py` — passes. Checked against two mutations:
  dropping the install's `|| echo` fails the "every step fails" case, and
  accepting any `freeze` on `PATH` fails the "freeze at another version" case.
- Real install into a fresh `HOME` against the network: exit 0, nothing on
  stderr, and `freeze --version` prints `freeze version v0.2.2`, so the
  comparison matches the binary rather than only the stub.
- `GOBIN=<missing>/deep go install github.com/charmbracelet/freeze@v0.2.2` exits
  0 and writes the binary, which is why the hook creates no directory itself.
- With a real `freeze` copied to `/usr/bin`, the first version of the test fails
  its "every step fails" case and the current one passes on the same machine,
  which is what the stub-only `PATH` is for.
- `.claude/scripts/check-frontmatter.py` — 14 documents carry their frontmatter.
- `.claude/scripts/check-commit-message.py --range origin/edge..HEAD` — clean.
- No Rust or `docs/` file changed, so the Cargo gates and `pnpm --dir docs
  build` were not run.

## Not covered

Installing `freeze` on a local checkout, deliberately: the hook stays behind its
`CLAUDE_CODE_REMOTE` guard because a local machine already carries the renderer,
which is the environment `plan.md:189-191` writes for. The container is the one
that starts without it.

The capture harness's fallback when `freeze` is absent, which the harness owns.

https://claude.ai/code/session_01Lf86zgAS8E5Xbt3J61sH4U


desertthunder:
Answering the first standard pass, which posts nothing, so its findings are recorded here with their outcomes. Disclosure: one session wrote this diff, ran that pass, and made these fixes, against the rule in `internal/models.md` that an implementer and a reviewer never share a run.

Pushed 9409f4f and 4add5f7.

- `medium` · nothing checked the report-and-continue contract — **fixed**. `.claude/hooks/session-start-test.py` runs the hook against stub `rustup`, `cargo`, `pnpm`, and `go`: the zero exit, each report, the empty stdout, and the `HOME`-unset path. It reads the pinned version back out of the stub's arguments instead of repeating it, so the version installed and the version looked for cannot drift. A `hooks` job runs it with no toolchain and no network. Checked against two mutations: dropping the install's `|| echo` fails the "every step fails" case, and restoring `command -v` fails the "freeze at another version" case.
- `low` · `command -v freeze` passed for any freeze on `PATH` — **fixed**. `freeze --version` is compared against the pin, which reports both a failed install behind an older binary and an unreachable one. Confirmed against the real binary, not only the stub: a live install prints `freeze version v0.2.2` and the hook stays silent.
- `low` · the install directory was unnamed and `/root` was a guess — **fixed**. The failure message names the directory, and without `HOME` the hook reports and installs nothing rather than writing to an invented path.
- `low` · dead `mkdir -p` — **fixed**, dropped. `GOBIN=<missing>/deep go install …` exits 0 and writes the binary, so `go install` creates it.
- `nit` · "Every step reports a failure to stderr" overstated the hook — **fixed**. `internal/thunderstorm.md` now says no step's failure fails the hook, and names the test that holds it.
- `nit` · the body's verification line cited a run that did not test what it claimed — **fixed**. The body is reworded and the claim is backed by the direct check above.

— claude-opus-5 · reasoning level unstated by the harness

---

desertthunder:
Adversarial pass. Reviewed at 2ce25e5 on #48, after the rebase onto `edge` at ae8c53b. Disclosure: one session wrote this diff and ran all three passes over it, against the rule in `internal/models.md`. Three findings, none of them a runtime failure; the first two come from the base moving under the issue.

```text
high · .claude/hooks/session-start.sh:42-44 — the comment's premise was reversed on edge one commit before this branch rebased onto it: internal/features/tui-verification/plan.md:187 now writes .ansi to ignored scratch and posts the frames to the pull request as stripped text, so "a committed .ansi capture" and "no image is committed" describe a workflow the plan deleted → a reader who follows the comment goes looking for internal/qa/captures/, and the same sentence carries into internal/thunderstorm.md:60-62 ("ANSI text nobody can look at"), which is now the text a reviewer does read → reword both against plan.md:187-193
medium · .claude/hooks/session-start.sh:10-12,41-73 — the hook installs freeze only when CLAUDE_CODE_REMOTE=true, while plan.md:189-191 calls freeze "a local convenience here rather than a step in the loop" and plan.md:214-217 records that no agent route uploads an image to a comment → the one environment that gets the renderer is the container whose images reach nobody and which is reclaimed at the end of the session, and a local checkout, where the plan says the image is looked at, never gets it; issue #30 predates ae8c53b and does not record which way to resolve it → escalation rather than an edit: either the install moves outside the cloud guard or the plan stops calling freeze local
low · .claude/hooks/session-start-test.py:68,100 — the "every step fails" case puts /usr/bin:/bin on the stub PATH and then asserts "is not on PATH (found nothing)" → a machine carrying a real freeze in /usr/bin fails that case for a reason unrelated to the hook, which is a test failing on the reviewer's environment rather than on the change → put only the stub directory on PATH and launch the hook through an absolute /bin/bash, since Python resolves the executable through the PATH it is handed
```

Checked and not reported: the install-then-probe sequence works in one shell run (a real install into a fresh `HOME` leaves the probe silent, so bash's command hashing does not hide the new binary); the version comparison matches the real binary's `freeze version v0.2.2` and not only the stub; both new files carry mode 100755, so the CI job can execute the test; the stale-version and dropped-`||` mutations both fail the suite.

— claude-opus-5 · reasoning level unstated by the harness

---

desertthunder:
Answering the adversarial pass (— claude-opus-5 · reasoning level unstated by the harness). Pushed 633dc57.

- `high` · the hook comment described committed captures — **fixed**. `.claude/hooks/session-start.sh:42-46` now says the harness writes `.ansi` to ignored scratch and posts the frames as stripped text, and points at `internal/features/tui-verification/plan.md` for the rest. `internal/thunderstorm.md` follows it.
- `medium` · freeze installs only in the cloud, while the plan calls it a local convenience — **deferred, escalated**. Resolving it means either moving the install outside the `CLAUDE_CODE_REMOTE` guard, which makes every local session pay a `go install` the hook has never paid, or changing `plan.md` to stop calling freeze local. Neither is recorded in #30, and the plan still names this hook as the install site, so this pull request does what the issue asks and the decision goes back to the author.
- `low` · the test left `/usr/bin` on the stub PATH — **fixed**. The stub directory is the whole PATH and bash is launched by a resolved path, since the hook calls no command the stubs do not supply. Confirmed the finding rather than assuming it: with a real freeze copied to `/usr/bin`, the previous test fails its "every step fails" case and the revised one passes on the same machine.

Both mutations still fail the suite after the change: dropping the install's `|| echo` fails "every step fails", and accepting any freeze on PATH fails "freeze at another version".

— claude-opus-5 · reasoning level unstated by the harness

---

desertthunder:
The escalated adversarial finding is closed by the author: a local checkout already carries `freeze`, so the `CLAUDE_CODE_REMOTE` guard stays and the container is the only environment that needs the install. No code change; the body's `## Not covered` records it.

— claude-opus-5 · reasoning level unstated by the harness

---

Issue #27, opened 2026-09-17
Fix the vocabulary a transcript row speaks


desertthunder:
From `01M2RP8M7WPM6D6SMA4TF6ZHKW`, `internal/features/transcript/plan.md`, over the seven `## UI` items in `internal/BUGS.md` (`01M2RFP6G4NBXT94SAAYZ1D1FH`).

## Goal

The seven `## UI` items land against one set of names rather than seven. A tool outcome is typed data on the contract and in the session record instead of prose the renderer parses back. An activity kind says what happened. `Detail` says how much the reader asked for, separately from the width-derived `Density`. The eleven row families stay eleven.

## Done when

`ActivityKind` has exactly the seven variants `Read`, `Search`, `Edit`, `Test`, `Lint`, `Build`, and `Command`; `ToolOutput` and `ToolFinished` carry exit status and elapsed time as typed fields; `shell_result_metadata` runs only as a named fallback for records written below the raised `schema_version`; a `Detail` type exists with `Compact`, `Normal`, and `Expanded` while `Density` keeps its name and its three width-derived variants; and `TranscriptRowKind` still has eleven variants. The `## UI` list in `internal/BUGS.md` holds no unchecked item that is not an open issue.

## Dispatch order and file ownership

#37 is the base: the typed outcome is what a kind, a diagnostic, and a live elapsed counter all read. Four sub-issues wait on it directly.

The dependencies are recorded as GitHub issue dependencies, so each sub-issue shows what blocks it and GitHub refuses to close one out of order. That relation is authoritative; the waves below are what it implies, together with the file ownership a run needs to decide what is safe to work at once.

| Wave | Issues | Owns |
| --- | --- | --- |
| 1 | #37 | `thndrs-agent/src/contracts.rs`, `core/tools/shell.rs`, `cli/renderer/view/transcript.rs` |
| 1 | #41 | `cli/app/input.rs` |
| 1 | #44 | `cli/renderer/tool_output.rs`, `cli/renderer/diff.rs` |
| 2 | #38 | `core/session/records.rs`, `core/session/reader.rs` |
| 2 | #39 | `cli/renderer/transcript.rs`, `cli/renderer/view/transcript.rs` |
| 2 | #40 | `cli/renderer/layout.rs`, the detail switch in `cli/app/` |
| 3 | #43 | `cli/renderer/view/transcript.rs`, the reporting tools |
[...]

PR #49, opened 2026-09-17
chore: generate capture fixtures from session records

### LINKED ISSUE #29: Generate session fixtures by serializing SessionRecord values
From `01M2RNSY07A522766KBV2DPR41`, `internal/features/tui-verification/plan.md`. Nothing blocks this. The capture harness in this epic waits on it.

The acceptance criteria below are unchanged by the epic's move from committed captures to captures posted on a pull request. Only the framing in this section changed.

## Problem

A frame is evidence about the interface only if the transcript behind it is fixed. A capture over a transcript that varies run to run shows a reviewer the transcript, so every scenario past `startup` loads a session fixture and no scenario sends a prompt. `--session-dir` is a top-level flag on `Cli` (`crates/thndrs/src/cli/mod.rs:277`) and `/resume <session-id>` already exists (`crates/thndrs/src/cli/app/commands.rs:80`), so a fixture directory plus a resume reaches a populated transcript with no model call. The fixture still has to come from somewhere, and the two obvious sources are both wrong.

Hand-written JSONL is brittle. `SessionRecord` is a serde-tagged enum whose every variant carries `schema_version`, a `seq` that must strictly increase, and an ISO 8601 `time`, and `SessionReader::read_validated_records` (`crates/thndrs/src/core/session/reader.rs:16`) rejects a corrupt line, an out-of-order sequence, or a first record whose `session_id` disagrees with the one being opened. Hand-written lines satisfy those rules until the enum changes, and then fail at capture time with no compiler warning first.

A recorded session is worse. It carries whatever was in the real transcript, so it needs a redaction check for credentials, absolute paths, and private content on every refresh.

A generator that serializes `SessionRecord` values avoids both. It compiles against the enum, so a schema change breaks the build rather than the capture, and it invents its content, so there is nothing to redact.

## Done when

A generator writes a fixture session directory holding one session per capture scenario past `startup`, built by constructing and serializing `SessionRecord` values rather than by emitting text. `SessionReader::read_validated_records` accepts every generated session, asserted by a test that runs the generator and reads back what it wrote.

The generator adds no user-visible command: `thndrs --help` and every subcommand's help output are unchanged, asserted by a test.

No `.jsonl` file is committed. The generator writes into scratch space that `.gitignore` covers, and `git status --porcelain` after a generation run reports nothing.

## Not in this issue

Whether the generator lives as an example, a development binary, or a test helper. That is an implementation choice the spec leaves open.

The tmux harness that invokes it, and the captures it feeds.

Typed tool outcomes in the records. `ToolFinished` persists `status` and `output` and nothing else today, and raising that is the transcript epic's work in #27. Fixtures written here carry what the record can carry now, and the next capture run after that lands shows the richer rows.

desertthunder:
Closes #29

## What

`session::fixtures` writes one session per capture scenario past `startup` by
constructing `SessionRecord` values and serializing them. Eleven sessions:
`picker-open`, `streaming-mid-tool`, `tool-output-truncated`,
`permission-prompt`, `error`, `narrow-60-cols`, `short-16-rows`, `no-color`,
and one per `Theme` variant, each named for its scenario so the `/resume`
argument and the capture name are the same word. Sequences, timestamps, and
content are fixed, so two runs produce the same bytes. The entry point is
`cargo run --example session_fixtures` — an example, not a subcommand, so no
help output moves.

## Why

#29. Hand-written JSONL satisfies `read_validated_records` until the record
enum changes and then fails at capture time with no compiler warning first; a
recorded session needs a redaction pass on every refresh. A generator compiled
against the enum has neither problem. `internal/features/tui-verification/plan.md`
fixes the scenario list this follows.

## Verification

- `cargo fmt`, `cargo clippy --workspace --all-targets --all-features --locked -- -D warnings`: clean.
- `cargo test --workspace --all-features --locked`: 1770 passed, 0 failed, 12 ignored, plus the four other targets green.
- 15 new tests in `session::fixtures`, including a validated read of every generated session, byte-for-byte repeat generation, and scenario coverage checked against `Theme::value_variants`.
- 2 new tests in `cli::tests` walk every command's long help and assert the top-level set; neither mentions fixtures.
- `cargo run --example session_fixtures` then `git status --porcelain`: no output.
- End to end through the app: `thndrs --session-dir target/tui-fixtures/sessions session list` lists all eleven, and `session show tool-output-truncated` replays the transcript.
- `pnpm --dir docs build`: 92 pages, links valid.

## Not covered

The tmux harness and the captures themselves (#26's remaining work). Nothing
resumes a fixture under a tty here — `session resume` needs one, and this
container has none, so that path is covered by the reader and transcript tests
rather than by a live run. `ToolFinished` still carries only `status` and
`output`; the fixtures say what the record can say today and get regenerated
when #27 lands. A resumed fixture is appended to by the run that opens it, so
the harness regenerates per pass; that is documented, not enforced.

https://claude.ai/code/session_01VTrRD3hsV1hdyHDkaqi9KW

---
_Generated by [Claude Code](https://claude.ai/code/session_01VTrRD3hsV1hdyHDkaqi9KW)_


desertthunder:
Edit pass on the first standard review of `001eedf7`, which posted nothing to this thread. That pass and this edit both ran in the session that wrote the branch. Fixed at `0c210b23`.

| Finding | Outcome |
| --- | --- |
| `medium` — `default_fixture_dir` was cwd-relative while `.gitignore` anchors `/target` at the workspace root, so generating from `crates/thndrs` left eleven files in `git status` | Fixed. The default resolves from `CARGO_MANIFEST_DIR`; `the_default_directory_is_ignored_scratch` now asserts the absolute path and fails against the old one, and a run from `crates/thndrs` writes to the workspace `target/`. |
| `medium` — the picker lists the run's own session above every fixture under a wall-clock id and activity time, so `picker-open` differs between runs | Deferred to #50, and named where the docs promised otherwise: the module header and the QA page both say the picker frame is not settled yet. |
| `low` — `pub mod fixtures` put a development generator in the published `thndrs_lib` surface, with no shipped consumer | Fixed. Behind `dev-fixtures`, with `required-features` on the example; `cargo build -p thndrs` no longer compiles it and the example refuses to run without the feature. |
| `low` — the writer lock name was duplicated from `SessionWriter` and the test rebuilt the same string by hand | Fixed. `the_lock_path_is_the_one_the_writer_takes` creates a real writer and asserts the path this module names exists while it lives and is gone after it drops. |
| `low` — `blocks().len() > 1` did not check that a transcript is populated | Fixed. The test asserts the block count each scenario replays to, ten for the base transcript. |
| `nit` — file times are unpinned, so an mtime-ordered listing varies by filesystem granularity | Not a defect here. The picker sorts on record-derived `last_activity` (`inventory.rs:246`), and no capture scenario renders the mtime-ordered listing. |
| `nit` — the QA snippet ran `./target/debug/thndrs` without a build step | Fixed. |
| `nit` — the help check greps for the literal `fixture` | Not a defect. It is a tripwire; `top_level_commands_are_the_published_set` beside it is the guard that names the whole surface. |

`cargo fmt`, `cargo clippy --workspace --all-targets --all-features --locked -- -D warnings`, and `cargo test --workspace --all-features --locked` (1771 passed, 0 failed, 12 ignored) are clean, `cargo test -p thndrs --lib` without features runs all 16 fixture tests, and `pnpm --dir docs build` reports 92 pages with valid links.

— claude-opus-5 · high

---

desertthunder:
Adversarial pass at `53bbbe4`, after the merge from `edge`. This diff and both earlier passes ran in the session that wrote the branch, so read the findings as such. Four, none repeating the standard pass.

`medium` · `crates/thndrs/src/core/session/fixtures.rs:857` — the test that guards the anchored default recomputes the implementation's own arithmetic: `default_fixture_dir` walks `CRATE_DEPTH_FROM_WORKSPACE_ROOT` ancestors of `CARGO_MANIFEST_DIR` (`:349`) and the test walks the same constant from the same root before comparing. Set that constant to `1`, or move the crate a level up, and both sides agree on `crates/target/tui-fixtures/sessions` while the suite stays green and the generator writes where `git status` reports it — the defect the standard pass fixed, reinstated invisibly. The `unwrap_or(manifest)` fallback at `:350` swallows the other half of it. → find the root by walking up to the `Cargo.toml` carrying `[workspace]`, which no constant can move, and let a wrong layout fail rather than resolve.

`medium` · `crates/thndrs/src/core/session/fixtures.rs:728` — `generation_repeats_byte_for_byte` runs both generations in one process inside one second, and `timestamps_are_fixed_and_iso_8601` (`:744`) pins only the first record's time, asserting a `T` and a trailing `Z` for every other. Change `FixtureBuilder::stamp` (`:201`) back to `datetime::now_iso8601()` and every test passes while the fixtures stop matching the ones a capture run took yesterday, which is the whole property this module exists for. → assert each scenario's last timestamp, a pure function of `FIXTURE_START_UNIX` and `FIXTURE_STEP_SECONDS`.

`low` · `crates/thndrs/src/core/session/fixtures.rs:347` — `CARGO_MANIFEST_DIR` is the build machine's path, so a dependent crate enabling `dev-fixtures` resolves the default to `~/.cargo/registry/src/target/tui-fixtures/sessions` and `generate` creates directories inside the cargo cache. Nothing reaches it today: the feature is development-only and the example is not packaged. Traced, not run. → either require the argument in a non-workspace build or return the root as a `Result`.

`nit` · `docs/src/content/docs/docs/development/tui-qa.md:48` — the snippet passes a relative `target/tui-fixtures/sessions` one paragraph after explaining that a relative path resolves against the shell's directory, so a reader who runs it from `crates/thndrs` gets exactly the behavior the anchoring prevents. → drop the argument from the snippet and let the default stand.

Nothing found on the hostile-input, concurrency, or invariant side beyond what the standard pass raised: scenario names are compile-time constants so no path escapes the fixture directory, ids share no prefix that `resolve_session_file` could confuse, and `read_validated_records` is asserted over every generated session rather than a sample.

— claude-opus-5 · high

---

desertthunder:
Edit pass on the adversarial review of `53bbbe4`. All four fixed at `16beb28`.

| Finding | Outcome |
| --- | --- |
| `medium` — the directory test walked the same ancestor count the implementation did, so a wrong count moved both sides together | Fixed. The root is now the nearest manifest carrying `[workspace]`, read at runtime, and `the_default_directory_is_one_git_ignores` asserts the `/target` rule in that root's `.gitignore` and the path's position under it. Pointing the resolution one directory up makes it fail. |
| `medium` — `generation_repeats_byte_for_byte` ran both generations inside one second and only the first record's time was pinned, so a wall clock would have passed | Fixed. `timestamps_are_the_same_on_any_day` asserts the literal first and last timestamp of every scenario and that the clock advances. Swapping `stamp` back to `datetime::now_iso8601()` fails it; I ran both sabotages rather than reasoning about them. |
| `low` — a dependent crate enabling `dev-fixtures` would resolve the default into the cargo registry cache | Fixed by the same change. A packaged manifest carries no `[workspace]` table — checked against `target/package/` — so there is no root above a registry copy and `default_fixture_dir` returns an error instead of a path. It returns `io::Result` now; the example propagates it. |
| `nit` — the QA snippet passed a relative path right after explaining why relative paths are the trap | Fixed. The snippet takes no argument and the paragraph says what a passed path resolves against. |

The branch also carries `edge` (`54f547d`), including the plan change in #51. The scenario names that change reads from are unchanged, so the eleven fixtures still match it; only which frames get posted moved, and that belongs to the harness.

CI: `cargo doc --workspace --all-features --no-deps --locked` under `-D warnings` failed on `53bbbe4`'s predecessor because an outer doc comment on the `mod` declaration merges with the file's `//!` header and resolves the whole of it in the parent's scope. Fixed at `53bbbe4`, and `cargo doc` has joined the local gate list. Every CI command now runs here before a push: fmt, clippy, test, doc, `pnpm --dir docs build`, `cargo package -p thndrs-agent`, and the frontmatter checks, all clean, with 1771 tests passing.

— claude-opus-5 · high

---

